PR #30, opened 2019-06-17
Adds the function for reading tiffs into minian.

eichlab:
To address a user issue requesting the ability to read tifs instead of avis.


minian/utilities.py
```diff
@@ -113,7 +113,16 @@ def load_videos(vpath,
...
+
+    file_extension = os.path.splitext(vlist[0])[1]
+    if file_extension == '.avi':
+        movie_load_func = load_avi_lazy
+    elif file_extension == '.tif':
+        movie_load_func = load_tif_lazy
+    else:
+        raise ValueError('Extension not supported.')
+
+    varr_list = [movie_load_func(v) for v in vlist]
```
phildong: I would be careful to follow this exact same flow for tif here. When I wrote this `v` was a single video, which contains 1000 frames, `load_avi_lazy` was designed to do a lazy loading in that case, then they are concatenated here. I see you almost copied that behavior for `load_tif_lazy`, which would make sense for several tif stacks that should be concatenated together, but not when each frame is a single tif file. I think it would produce significant overhead in that case or no longer doing a lazy load at all (since there is a `compute` of the first frame in `load_tif_lazy`). Have you tested this with a folder of single-frame tifs?

phildong:
Thanks Will! I see you incorporated changes from both `dev` and `mc` branch. I don't think we are convinced stuff under `mc` branch are the way to go, and they definitely need to be cleaned up before merging into actual developments. Do you mind taking those out? Also I think it would make more sense to merge this into `dev` instead of `master` now since a lot of stuff under `dev` are not ready either.

PR #31, opened 2019-06-24
Zdev

ZachPenn:
Zach's edits to minian_noted.ipynb and fixes to make demo movie analyzable.


ZachPenn:
> I would say `if len(bad_units) > 0` to be more explicit, then `else` later

This has now been corrected under commit c5794ec for both minian.ipynb and minian_noted.ipynb

pipeline.ipynb
```diff
@@ -64,40 +73,56 @@
    },
    "outputs": [],
    "source": [
+    "#Set up Initial Basic Parameters#\n",
     "minian_path = \".\"\n",
-    "dpath = \"./demo_movies\"\n",
+    "dpath = \"./demo_movies/\"\n",
```
phildong: is this intended?

pipeline.ipynb
```diff
@@ -146,13 +175,13 @@
...
     "    'dl_wnd': 5,\n",
-    "    'sparse_penal': 0.05,\n",
+    "    'sparse_penal': 0.1,\n",
     "    'update_background': False,\n",
     "    'post_scal': True,\n",
     "    'zero_thres': 'eps'}\n",
     "param_second_temporal = {\n",
     "    'noise_freq': 0.06,\n",
-    "    'sparse_penal': 1,\n",
+    "    'sparse_penal': 20,\n",
```
phildong: not sure that's a good choice for most of the data out there. If this is specific to certain data let's not changing the parameters back-and-forth

pipeline.ipynb
```diff
@@ -520,9 +538,9 @@
    "source": [
     "%%opts Curve [width=500, tools=['hover']]\n",
     "hv.output(size=output_size)\n",
-    "if interactive and in_memory:\n",
-    "    hv.NdOverlay(dict(width=hv.Curve(shifts.sel(variable='width')),\n",
-    "                      height=hv.Curve(shifts.sel(variable='height'))))"
+    "if interactive: #and in_memory:\n",
```
phildong: `in_memory` is gone in `dev` and you can safely take it out forever

pipeline.ipynb
ZachPenn: No, this was by accident and was fixed in commit `a6727ba`

pipeline.ipynb
ZachPenn: This was changed back to `sparse_penal : 1` in commit`a6727ba`

pipeline.ipynb
```diff
@@ -64,40 +73,56 @@
    },
    "outputs": [],
    "source": [
+    "#Set up Initial Basic Parameters#\n",
     "minian_path = \".\"\n",
-    "dpath = \"./demo_movies\"\n",
+    "dpath = \"./demo_movies/\"\n",
     "chunks = {\"frame\": 1000, \"height\": 50, \"width\": 50, \"unit_id\": 100}\n",
-    "subset = None\n",
+    "subset = dict(frame=slice(0,1000))\n",
```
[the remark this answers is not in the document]
ZachPenn: I don't think `slice(0,1000)` is a good default, but I do think it would be helpful to have a slice object that makes it clearer how to subset the data.  I have changed it to `subset = dict(frame=slice(0,None))` in commit `a6727ba`

pipeline.ipynb
ZachPenn: This has been removed in commit 4b9997b

PR #32, opened 2019-06-28
Adds the function for reading multi-page tiffs into minian

wmau:
Currently only handles tiff files that contain multiple frames. Single-frame tiff files will probably produce significant overhead.


phildong:
it looks quite a few things are not up-to-date with the dev branch and that result in a lot of conflicts (revert back), do you mind doing another merge from dev? it should be a relatively easy merge

PR #42, opened 2019-07-04
Save B_mrg_it2 as bl and update reference in fx generate_videos

ZachPenn:
`B_mrg_it2` was saved as `B`, which was in conflict with `b_spatial_it2` being saved as `b`.  `B` was being overwritten by `b`.  Changed saving of `B_mrg_it2` to `bl` in order to avoid this issue and amended code in fx `generate_videos` to reflect this change.


phildong:
I'm not sure I understand the problem. What do you mean by 

> `B` was being overwritten by `b` ?

At least on my machine they are saved as different variables with different names and dimensions.

ZachPenn:
It seems this is an issue with OSX, which does not by default use case sensitive filenames.  Also, I made it so `bl` would be `BL` so that the letter 'l' would not be confused with number '1'.

PR #46, opened 2019-11-27
Significant speed improvement during initial movie load

wmau:


phildong:
since this is more an improvement than bug fix let's merge this into dev branch first.

PR #47, opened 2020-01-28
Permit loading mkv videos as well

ximion:
Hi!
Thank you for creating Minian! Our [PoMiDAQ](https://github.com/bothlab/pomidaq) data acquisition software generates FFV1-encoded Matroska video files by default (to greatly reduce the size of stored videos while not loosing quality and using patent-free open-source codecs). Minian loads these just fine, and data analysis seems to work well in preliminary tests with short sample data.

So, we are using the attached patch to load MKV videos as well.
Cheers,
    Matthias

### DIFF
--- a/minian/utilities.py
+++ b/minian/utilities.py
@@ -116,7 +116,7 @@ def load_videos(vpath,
     print("loading {} videos in folder {}".format(len(vlist), vpath))
 
     file_extension = os.path.splitext(vlist[0])[1]
-    if file_extension == '.avi':
+    if file_extension in ('.avi', '.mkv'):
         movie_load_func = load_avi_lazy
     elif file_extension == '.tif':
         movie_load_func = load_tif_lazy


phildong:
Wow I'm surprised opencv can load it like this without any issue.
Thank you so much for your contribution!

ximion:
OpenCV can open pretty much anything FFmpeg can, which is *a lot* of stuff. Since MKV is super popular in general (mostly via it's WebM variant) and [FFV1](https://en.wikipedia.org/wiki/FFV1) is also very common for data archiving, this shouldn't cause an issue for most software which handles video.
The pixel format conversion to OpenCV BGR matrices may be expensive for some formats though, so while everything loads, some data may load slower.

Thanks for merging the PR!
We are currently evaluating Minian vs MIN1PIPE in two labs in our institute, and comparing them is actually a lot harder than initially thought. In any case though, Minian's interactive notebook is an excellent way to learn how the algorithms work and how they change the data, thank you a lot for writing the documentation!

PR #72, opened 2020-11-02
"Global" noise frequency variable

wmau:
noise_freq appears at multiple steps including seed initialization and temporal update. Use a common noise_freq variable to ensure that this value is consistent across all these steps.


pipeline.ipynb
```diff
@@ -1192,7 +1193,7 @@
     "    p_ls = [1]\n",
     "    sprs_ls = [0.1, 0.5, 1, 2]\n",
     "    add_ls = [20]\n",
-    "    noise_ls = [0.06]\n",
+    "    noise_ls = [noise_freq]\n",
```
phildong, line 1196: can you restore this to numbers?

pipeline.ipynb
```diff
@@ -1647,7 +1648,7 @@
...
```
phildong, line 1651: can you restore this too?

pipeline.ipynb
```diff
@@ -1985,7 +1986,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.1"
+   "version": "3.8.6"
```
phildong, line 1989: can you take this change out?

phildong (changes requested):
could you apply the same change to `pipeline_noted.ipynb` as well?

PR #79, opened 2020-11-17
Meta cell dev removed toy dataset

zsinnema:
the toy dataset isn't used anywhere so I removed it


zsinnema:
this one is based on MetaCell-dev-add-test-fixture PR #78

PR #88, opened 2020-11-25
Perf

phildong:
merge performance branch


zsinnema:
@phildong can you please run black to fix code style errors? this will fix the travis build
btw I pushed a commit to disable running the tests and upload the coverage in the travis build. the reason for this is that the 50 minutes limit for travis build is not enough to finish running all tests
you can run the tests on your local machine and upload the coverage manually

PR #96, opened 2021-01-11
Global noise_freq

wmau:
All references to noise frequency now reference a common variable


phildong:
As discussed I think some media filter are more appropriate than the `detrend` function since it apply better to cases that cannot be fitted by a line
Also scipy functions usually do not release GIL and might not be very friendly to parallel processing

phildong:
closing since it's out-dated and has conflicts now.

PR #103, opened 2021-03-15
Annotation throttling

phildong:
use dask annotation throttling and graph optimization to regulate memory usage


environment.yml
```diff
@@ -1,16 +1,17 @@
 name: minian
 channels:
   - conda-forge
+  - simpleitk
```
zsinnema, line 4: @phildong can you please also update the requirements-base.txt in the requirements folder?
this is needed to install the MiniAn package through pip

PR #176, opened 2022-01-25
Pinning FFTW version

slarson:
Experienced an issue after setup when 0.13.0 of pyFFTW was released.  If we pin to 0.12.0 this seems like it won't happen

### DIFF
--- a/requirements/requirements-base.txt
+++ b/requirements/requirements-base.txt
@@ -16,7 +16,7 @@ numpy==1.20.2
 opencv-python==4.2.0.34
 pandas==1.2.3
 panel==0.8.0
-pyfftw>=0.12.0
+pyfftw==0.12.0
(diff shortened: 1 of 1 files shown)


phildong:
looks like a change in `environment.yml` is also needed.
Plus it's interesting to see it seem to only affect linux...

phildong:
@slarson if you could amend the `environment.yml` or give me permission to push to your branch it would be awesome!
Otherwise I might make a new PR soon.

slarson:
Thanks @phildong .  I'm pretty sure you have that permission already (See screenshot).  I'm not in the best position to ensure the build isn't broken so I think you had better take that on.  This PR was just intended to get the ball rolling on the observation that the build had been broken by a new release of pyFFTW.

Screen Shot 2022-02-07 at 7 29 12 AM

PR #193, opened 2022-03-15
Get correct frame amount for video files without nb_frames property & allow codec selection

ximion:
Hi!
First of all, I'm really impressed by this project! Two years ago, it was a bit mediocre compared to MIN1PIPE, but now based on very early testing, Minian seems to outperform CaImAn and MIN1PIPE for my datasets! (Still need a lot more testing for that though, I wonder whether switching out Minian's median filter with an anisotropic filter for denoising would make things even better...)
The best thing is that Minian's documentation is excellent, which makes it very easy to get started, and all public API seems to be well documented as well. Really awesome work!

The data I generate is recorded with [PoMiDAQ](https://github.com/bothlab/pomidaq) or [Syntalos](https://github.com/bothlab/syntalos), which prefer the MKV container for video, as it's a bit nicer to store FFV1 video in, but more importantly I use MKV's ability to save arbitrary metadata with the video file quite excessively.
When loading these files into Minian, it currently crashes as the `nb_frames` property apparently doesn't exist for MKV containers. Fortunately, the amount of frames is very easy to calculate, so this patch adds support for just that!

Thanks for considering the change! :-)
Cheers,
    Matthias

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,4 @@ demo_movies export-ignore
 .readthedocs.yaml export-ignore
 codecov.yml export-ignore
 .github export-ignore
+demo_movies/*.mkv filter=lfs diff=lfs merge=lfs -text
--- a/demo_movies/sample1.mkv
+++ b/demo_movies/sample1.mkv
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:ebea009f040f900b59732552616805e9a6e9c4ddf173d26a10ac4d30f388ef70
+size 156097649
--- a/minian/utilities.py
+++ b/minian/utilities.py
@@ -225,7 +225,15 @@ def load_avi_lazy(fname: str) -> darr.array:
     video_info = next(s for s in probe["streams"] if s["codec_type"] == "video")
     w = int(video_info["width"])
     h = int(video_info["height"])
-    f = int(video_info["nb_frames"])
+    if "nb_frames" in video_info:
+        f = int(video_info["nb_frames"])
+    else:
+        r_fps = video_info["r_frame_rate"]
+        num, denom = r_fps.split("/")
+        fps = float(num) / float(denom)
+        f = float(probe["format"]["duration"]) * fps
+        f = np.round(f).astype(int)
+
     return da.array.from_delayed(
         da.delayed(load_avi_ffmpeg)(fname, h, w, f), dtype=np.uint8, shape=(f, h, w)
     )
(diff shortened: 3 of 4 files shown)


ximion:
This PR also includes a change to allow setting the video codec. When depending on `ffmpeg~=5.0` (and using [Mamba](https://github.com/mamba-org/mamba) to resolve the environment in seconds, as conda takes almost a day and doesn't find a good solution) I can do this now using [AV1](https://en.wikipedia.org/wiki/AV1):
```python
generate_videos(varr.sel(subset),
                Y_fm_chk,
                A=A, C=C_chk,
                vpath=sys_tmpdir,
                vname='overview.mkv',
                vcodec='libsvtav1',
                options={'preset': '4', 'g': '240'},
                verbose=True)
```
Using AV1, the 1.4GB overview video is now just 2.2MB in size! This makes it possible to just always generate this video as part of a batch processing pipeline and just keep it around, while with the larger video handling the data would be more of a pain (and the backup costs would explode in the long run). The AV1 quality with these settings is a bit worse than H.246, but at least for me that's definitely acceptable here :-)

ximion:
Test failures look unrelated to this PR.

phildong:
Hey @ximion  Thank you so much for the PR! I skimmed through the changes and everything looks great!
I'm wondering whether you can provide a small piece of mkv data to be stored on the minian github so that we may be able to incorporate this in part of the test?
At the same time I'll try to resolve the testing failure for now

ximion:
Hi @phildong and sorry for the delay! Congrats on publishing the Minian paper! :-)

> I'm wondering whether you can provide a small piece of mkv data to be stored on the minian github so that we may be able to incorporate this in part of the test?

Sure, I added a small 30sec piece (which is still 148MB in size) to this PR, let me know if you need a longer slice.
I did enable [Git LFS](https://git-lfs.github.com/) for this file, to store the large binary file in this Git repository efficiently.
It may make a lot of sense to enable this for the AVI files as well, but to do that you would need to rewrite the Git history (so only makes sense after merging this PR, and only if you want to).

The commands for a LFS conversion of this repository (after making sure the LFS module is installed as described at https://git-lfs.github.com/) would be:
```bash
# track .avi files in LFS
git lfs track "demo_movies/*.avi"

# test what a migration run would do to the repo
git lfs migrate info --everything --include="demo_movies/*.avi"

# actually migrate everything
git lfs migrate import --everything --include="demo_movies/*.avi" --verbose

# force-push, possibly rewriting the history
git push -f

# ensure working copies of the large files are up to date
git lfs checkout
```
Sidenote: I don't think LFS is an amazing solution in every case, as it stores the large objects on a central server, breaking Git's decentralized approach, it adds an extra step to using a repo and it is annoying to switch away from it on a repo that is using it. However, for large test data as is the case with this, I think its benefits definitely outweigh the cost at the moment :-)

ximion:
Any news on this? We have been using this code for pretty much a year now without any issues :-)

ximion:
Closing this in favour of https://github.com/denisecailab/minian/pull/246, since I need the master branch of the original fork for, well, our Minian fork ^^
(I didn't expect that this would take so long to get merged)

PR #205, opened 2022-05-24
updated update_meta()

wmau:
New update_meta() function can now permanently change saved metadata.


minian/utilities.py
```diff
@@ -624,30 +624,59 @@ def xrconcat_recursive(var: dict | list, dims: list[str]) -> xr.Dataset:
...
-            if backend == "netcdf":
-                new_ds.to_netcdf(f_path, mode="a")
-            elif backend == "zarr":
-                new_ds.to_zarr(f_path, mode="w")
+
+            # Save each variable back to its own zarr store. Write to a temp
+            # store first and swap, since the variables are lazily backed by the
+            # very stores we would otherwise overwrite while still reading them.
+            for varname, da in ds.data_vars.items():
+                store = os.path.join(f_path, varname + ".zarr")
```
sneakers-the-rat: dropping support for netcdf?

minian/utilities.py
```diff
@@ -624,30 +624,59 @@ def xrconcat_recursive(var: dict | list, dims: list[str]) -> xr.Dataset:
...
+                tmp_store = store + ".tmp"
+                var_ds = da.to_dataset()
+                var_ds.attrs = deepcopy(ds.attrs)
+                var_ds.to_zarr(tmp_store, mode="w")
+                shutil.rmtree(store)
+                os.replace(tmp_store, store)
```
sneakers-the-rat: surely there must be some way to update attributes without needing to copy the entire thing, right?

minian/utilities.py
daharoni: Looks like netcdf was already removed from `save_minian` and `open_minian` on thane current master branch. So I think netcdf in `update_meta` is a dead on master and in our modernizer-minian branches so I removed it.

Likely not that useful but we could add in a small netcdf exporter but I would lean to just dropping netcdf for now.

minian/utilities.py
daharoni: Right! Fixed in https://github.com/miniscope/minian/pull/205/commits/f1279d4dec5d81b9f4c8ffbac0f3dd0bd9174513 where we just append the coordinates.

minian/utilities.py
sneakers-the-rat: yeah same, in favor of just having zarr be the format. https://github.com/miniscope/minian/issues/319

minian/test/test_utilities.py
```diff
@@ -0,0 +1,116 @@
...
+    update_meta(str(tmp_path), meta_dict={"session": -1, "animal": -2})
+
+    after = open_minian(str(mn_path))
+    assert str(after.coords["session"].values) == "session1"
+    assert str(after.coords["animal"].values) == "animalA"
+    # The underlying data must be untouched by the metadata update.
+    np.testing.assert_array_equal(after["test_var"].values, var.values)
+
+
+def test_update_meta_is_lazy(tmp_path):
```
sneakers-the-rat, line 46: just added one quick test that directly tests the non-updated data isn't touched.

daharoni: nice. We good to merge?

daharoni (approved):

sneakers-the-rat (approved):
hell ya.

We haven't really talked about deprecation warnings, and this is one that seems like it can skate by, but for stuff that's in the main pipeline we should emit warnings before breaking changes (probably just for one version, in 1.3.0 and then can be removed in 2.0)

PR #253, opened 2023-11-07
add logo to readme

bendichter:


sneakers-the-rat (approved):
what up ben, reviving this package. lgtm :)

actually what the heck where is that image in the repo

PR #294, opened 2025-07-24
Python >=3.10 compat, update packaging

sneakers-the-rat:
@phildong got this started for ya last night, basically i just updated the packaging to modern pyproject.toml style, set a new minimum, ran ruff's pyupgrade formatter, and then started running the tests and whacking away at whatever bugs started to happen. mostly just simple API changes in dependencies. shouldn't take too much longer to get it working. bonus: this should also resolve many of the installation issues i'm seeing as well as make it possible to put this up on pypi.

opening as a draft with allow edits by maintainers, you should be able to push directly to this branch like

```
git remote add sneakers-the-rat https://github.com/sneakers-the-rat/minian
git fetch sneakers-the-rat
git switch -c py39 sneakers-the-rat/py39
# do stuff... commit... push as normal
```


minian/__init__.py
```diff
@@ -1,8 +1,12 @@
 import dask as da
 import os
+from importlib.metadata import version
 from .utilities import custom_arr_optimize, custom_delay_optimize
 
-__version__ = "1.2.1"
+try:
+    __version__ = version("minian")
```
sneakers-the-rat, line 7: wasn't sure why version was defined in 4 different places, so reduced it to only two (not sure what that `committizen` tool does)

pipeline.ipynb
daharoni: This is throwing an error due to a change in IPython. The fix is to change this to `from IPython.display import display` for new versions of IPython

daharoni:
pymetis is only used in 1 place and looks like `from networkx.algorithms.community import greedy_modularity_communities` can achieve basically the same thing. Going to add a fall back to this for the cases where pymetis fails to import.

daharoni:
Another fix incoming is newer dask versions are more strick in their workers going over their mem limit. This causes the demo pipeline notebook running on the demo dataset to end up erroring out.

Going to default the per worker mem limit to 4GB instead of 2GB and also add a path for setting this as an environment variable.

minian/cnmf.py
```diff
@@ -35,6 +33,27 @@
...
 )
 
+try:
+    import pymetis as _pymetis
+
+    _HAS_PYMETIS = True
+except ImportError:
+    # pymetis is the gold-standard k-way graph-partition library but its
+    # only build artifact is from source — no Windows wheels on PyPI for
+    # any version. Fall back to a pure-Python NetworkX partition below.
```
sneakers-the-rat:
```suggestion
    # pymetis doesn't build windows wheels - fallback to nx if not present
```

i just sort of can't stand LLM comments, style recommendation

minian/cnmf.py
```diff
@@ -1858,13 +1883,43 @@ def adj_list(G: nx.Graph) -> List[np.ndarray]:
...
     """
     gdict = nx.to_dict_of_dicts(G)
     return [np.array(list(gdict[k].keys())) for k in sorted(gdict.keys())]
 
 
+def _partition_graph(G: nx.Graph, n_parts: int) -> list[int]:
+    """
+    Partition `G` into `n_parts` chunks, returning a per-node partition index.
+
+    Uses pymetis (min-edge-cut, gold standard) when available. Otherwise falls
```
sneakers-the-rat:
```suggestion
    Uses pymetis when available. Otherwise falls
```

minian/cnmf.py
```diff
@@ -1858,13 +1883,43 @@ def adj_list(G: nx.Graph) -> List[np.ndarray]:
...
+    back to NetworkX's greedy_modularity_communities — pure Python, ships
```
sneakers-the-rat:
```suggestion
    back to :func:`~networkx.algorithms.community.greedy_modularity_communities` — pure Python, ships
```

if this project built proper docs we could link directly to the other proper docs:
if we are going to maintain this thing, docs are a thing we should do.

minian/cnmf.py
```diff
@@ -1858,13 +1883,43 @@ def adj_list(G: nx.Graph) -> List[np.ndarray]:
...
+    with networkx, slower for very large graphs but lets minian install on
+    platforms without a pymetis wheel (notably Windows).
+
+    The partition quality affects only computation chunking performance in
+    `graph_optimize_corr`, not result correctness.
```
sneakers-the-rat:
```suggestion
    back to NetworkX's greedy_modularity_communities -- slower, but pure python.
```

none of this is relevant to a human user

minian/cnmf.py
```diff
@@ -1858,13 +1883,43 @@ def adj_list(G: nx.Graph) -> List[np.ndarray]:
...
+    """
+    if _HAS_PYMETIS:
+        _, membership = _pymetis.part_graph(n_parts, adjacency=adj_list(G))
+        return membership
+    from networkx.algorithms.community import greedy_modularity_communities
+
+    nodes_sorted = sorted(G.nodes)
+    if n_parts <= 1:
+        return [0] * len(nodes_sorted)
```
sneakers-the-rat: this looks generic across implementations - put this first

minian/cnmf.py
```diff
@@ -1858,13 +1883,43 @@ def adj_list(G: nx.Graph) -> List[np.ndarray]:
...
+    try:
+        comms = list(greedy_modularity_communities(G, best_n=n_parts))
+    except (ValueError, nx.NetworkXError):
+        # Asking for more partitions than the graph can yield. Fall back
+        # further to connected components — every node still gets a label.
+        comms = list(nx.connected_components(G))
```
sneakers-the-rat: undocumented fallback, generic error catching condition, implicit behavior that deviates from caller intent ("we requested n partitions, but sometimes we just sort of get a random amount of partitions for some reason"). what does pymetis do? behavior should be consistent: if a request is logically invalid, that should throw, not be masked.

minian/cnmf.py
daharoni: Minian does generate this, https://minian.readthedocs.io/en/stable/, I think. So we have a starting point.

minian/cnmf.py
```diff
@@ -11,6 +11,7 @@
 import numpy as np
 import pandas as pd
 import pyfftw.interfaces.numpy_fft as numpy_fft
+import pymetis
```
sneakers-the-rat, line 14: this wouldn't really work if we have the dependency be optional. we would still need the import check and to just raise an error in the function that it's used in. wouldn't be able to import this module unless one installed `minian[performance]`

pyproject.toml
```diff
@@ -59,13 +60,6 @@ minian-install = "minian.install:main"
 numba = [
     "numba>=0.52.0",
 ]
-# Faster graph partitioning in graph_optimize_corr. Falls back to a pure-Python
-# NetworkX community-detection routine when pymetis isn't available — same
-# correctness, slower for very large graphs. Optional because pymetis has no
-# Windows wheels on PyPI and the from-source build requires C tooling.
-performance = [
```
sneakers-the-rat: optional style note: two schools of thought on naming optional dependency groups: by feature (docs, gui, plotting, etc.) or if the dependencies have a specific name that amounts to a feature switch (dask, numba, mypy). naming an optional dep group "performance" amkes me go "sure i will install the performance version why not" but doesn't really make it clear that not having the performance group makes a graph partitioning operation unavailable (since, from what i understand, the nx alternative doesn't have the same behavior and thus doesn't serve as an alternative).

minian/cnmf.py
sneakers-the-rat: ope i was looking at an old commit i suppose

sneakers-the-rat (commented):
i can't approve my own PR but looks good to me

pyproject.toml
daharoni: True. Buse just to add a follow up note for the future when/if we deal with pymetis, the nx alternative did still work, it just doesn't do a good jobat property grouping the footprint seeds. But this effect would only be seen at like 10,000+ seeds and maybe not even then.

So pymetis is a performance boost but only saves like a few seconds on normal sized data. The actual output of the resulting computation doesn't depend on how footprint/seeds are grouped, just saves on some memory and a few seconds compared to not doing any grouping at all.

requirements/ci-constraints.txt
sneakers-the-rat, line 1: this doesn't make a lot of sense to me to do, this is what the lockfile does, and we should set version caps that are appropriate to what the package can actually use in the pyproject.toml. having a test environment that differs from both the lockfile and the dependency specification in pyproject.toml means that the test environment doesn't reflect the deployment environment -> doesn't test if the package works.

daharoni: Yes I agree. I am just trying to get things to a point where CI is green so then I can work back and fix the things that were breaking it. There are just a bunch of small changes across the dependencies over the past 4 years that are resulting in weird behavior. 

And for the windows thing, I am 99% sure I can just drop pymetis without any downsides at all so was going to put effort into that rather than getting windows CI to run by installing deps differently than what is needed for linux.

.github/workflows/docs.yml
sneakers-the-rat, line 1: split this out to a separate workflow since it isn't related to testing, but i'm not sure why this is needed? typically the way you do readthedocs builds is by just having the .readthedocs.yaml file and configuring readthedocs to use github's webhooks. https://docs.readthedocs.com/platform/stable/intro/add-project.html

requirements/ci-constraints.txt
sneakers-the-rat: ok no problem. if pymetis can be dropped, that is certainly simplifying.

daharoni:
Ok, looks like the main issue with fully up-to-date deps is minian has some code that was a workaround of dask issues from a few years ago. Since then dask has addressed them but this breaks minian's past workarounds. Working on fixing this now.

Issue #302, opened 2026-05-29
Correlation axis bug in graph_optimize_corr


daharoni:
Found this while replacing pymetis with a built-in spatial partitioner, wrote a test that compared adj_corr output against np.corrcoef and they didn't match. Turns out it's a real bug, present since probably the original release.

What's happening:

Inside `graph_optimize_corr.construct_comput`, vsub comes out of xarray's vectorized indexing with shape (pixels, frame) which is the orientation idx_corr expects. The function then does:

```
if len(idx_arr) > 1:
    vsub = vsub.T
```

That transpose flips it to (frame, pixels). `idx_corr` iterates `X.shape[0]` for mean-centering and treats `ridx[k] / cidx[k]` as row indices, so it ends up computing the Pearson correlation between frame slices across the chunk's pixel subset rather than between the two pixel timeseries the caller asked for. The returned value also depends on which other pixels happened to share the chunk, so it isn't even chunk-invariant.

What is likely affected:

* seeds_merge: uses nod_df = seeds[["height","width"]], two-dim indexing -> transpose applied -> affected.
* initA: uses default idx_dims=["height","width"] -> affected.
* unit_merge: uses nod_df = pd.DataFrame({"unit_id": ...}), single-dim indexing → takes the else (rechunk(-1)) branch, no transpose -> not affected.

Why it wasn't caught:

This bug was introduced on 2021-04-05 in #103 ("Annotation throttling") as part of the dask-throttling refactor that also added the pymetis dependency. Present in v1.0.0rc0 and every release since.

The bug's accidental behaviour happens to overlap with what we wanted most of the time. Seeds are added to the graph in row-major order, so consecutive seed_ids tend to be spatially adjacent; the bug returns corr(frame[seed_a_id], frame[seed_b_id]) which is typically high for nearby frames of calcium imaging data; the radius-neighbor graph only proposes spatially adjacent pairs. Net effect: "merge spatially-close seeds whose IDs happen to be close", which approximates the correct rule ("merge spatially-close seeds with similar traces") enough that the figures look reasonable.

Empirical impact on the demo movie:

Ran seeds_merge twice on the demo data. Once with the bug in place, once with the fix. Same input seeds, comparing the resulting mask_mrg:

* Input seeds: 716
* Surviving merge (buggy): 322
* Surviving merge (fixed): 245
* Per-seed agreement: 631 of 716 (88%)
[...]

PR #303, opened 2026-05-29
Replace pymetis with k-d tree spatial partitioner

daharoni:
Replaces the `pymetis` dependency (no Windows PyPI wheel, MSVC + METIS C-lib needed from source) with a position-aware k-d tree spatial partitioner built into `minian.cnmf`. The graphs that `graph_optimize_corr` operates on only carry edges between geometrically close nodes, so a spatial split balances node counts and minimises cross-partition edges in one step.

## Commits

1. `fix(visualization): VArrayViewer box draw uses Rectangles` — Bokeh 3.x compat fix surfaced while testing the partition viz interactively. Reviewable in isolation.
2. `feat: replace pymetis with k-d tree spatial partitioner` — `spatial_partition`, `partition_diagnostics`, `chunk` kwarg plumbed through `adj_corr` / `seeds_merge` / `unit_merge` / `initA`, ValueError contracts, JSON-safe membership, Windows back in CI, deps cleaned, test goldens unchanged from master.
3. `feat(viz): partition preview visualization + notebook integration` — `visualize_spatial_partition` and a preview cell before `seeds_merge`.

(The originally-included `fix(cnmf): correct axis orientation in graph_optimize_corr` commit and its follow-ups have been reverted — see "Issue #302 retraction" below.)

## Issue #302 retraction

The PR initially also included a "fix" for issue #302, claiming `graph_optimize_corr.construct_comput` was transposing `vsub` the wrong way. That diagnosis turned out to be wrong; the master code was correct. The mechanism:

`xarray.DataArray.sel` with shared-dim indexers places the new dim at the position of the first replaced dim. So:

* `(frame, height, width).sel(height=DA, width=DA)` → `(frame, pixels)` = `(T, N)`
* `(height, width, frame).sel(height=DA, width=DA)` → `(pixels, frame)` = `(N, T)`

The pipeline produces `Y_hw_chk` as `(frame, height, width)` (`load_videos` writes that order; `apply_transform` and `save_minian` preserve it), so master's `if len(idx_arr) > 1: vsub = vsub.T` was correctly producing `(N, T)` for `idx_corr`. The synthetic test fixture authored alongside the "fix" used `(height, width, frame)` — the only orientation under which removing the transpose looks correct, and the only orientation under which `np.corrcoef` agreed with the without-`.T` code.

Empirical reproduction: on a `(frame, height, width)` varr, the without-`.T` code returns Pearson values that disagree with `np.corrcoef` by up to ~1.0; with the `.T` restored, the agreement is to 1e-7. The fixture has now been re-oriented to match production, so `test_adj_corr_returns_true_pearson` actually pins the production code path going forward.

Demo-data goldens accordingly revert to the master values (`unit_id ≈ 286`, `C.sum ≈ 546290`, `S.sum ≈ 5065`, `A.sum ≈ 71468`). `max_proj` and `motion` were upstream of `adj_corr` and were never affected.

## What's still in scope

* `spatial_partition(positions, target_chunk)` — recursive k-d tree median split on 2D node positions; replaces the `pymetis.part_graph` call that required a compiled METIS dependency.
* `partition_diagnostics(membership, adj=None, n_frames=None)` — sizes, intra/inter-partition edges, per-partition memory estimate.
* `chunk` kwarg on `adj_corr` / `seeds_merge` / `unit_merge` / `initA` so users can tune partition size without editing source.
* `visualize_spatial_partition` and a preview cell before `seeds_merge` for interactive `chunk` tuning.
* `unit_merge` now raises `ValueError` if any footprint has zero mass, naming the offending `unit_id` — matches `update_spatial`'s own contract that empty footprints are dropped upstream.
* Windows back in the CI matrix (`ffmpeg` via `choco`; `pymetis` removed).
* Packaging modernised: `pyproject.toml` + `pdm.lock` replace `setup.py` + `environment.yml`.

## Test changes

* New `minian/test/test_cnmf.py` with per-function unit tests for the new k-d tree partitioner, partition diagnostics, chunk-kwarg plumbing, and contract errors. The `synthetic_varr` fixture is now in production dim order `(frame, height, width)`.
* `test_cross_reg.py`: portable `sys.executable -m jupyter` subprocess invocation.
* `test_pipeline.py`: golden values unchanged from master (within widened tolerances that account for library-version drift on the locked snapshot).

## CI

Windows-latest restored to the matrix; first-and-last-Python-only exclusion pattern matches macOS.

Issue #302 closed as "not a bug — master was correct; original synthetic test was authored against the wrong dim order."


minian/test/test_cnmf.py
```diff
@@ -0,0 +1,422 @@
...
+            "width": np.arange(width),
+            "frame": np.arange(frames),
+        },
+    )
+
+
+class TestAdjCorrPositionPlumbing:
+    """The `positions` kwarg affects chunking only, not correlation values."""
+
+    def _setup(self, n=80, h=15, w=15, t=300, radius=4, seed=0):
```
sneakers-the-rat: one of the reasons i typically like to favor functional tests vs. class based tests is that they force you to write using fixtures, and this is something that would work well as a fixture that encapsulates setup and can handle scoping

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,422 @@
...
+    def test_with_smoothing(self):
+        # `freq` triggers the FFT lowpass branch; partition-invariance still
+        # holds.
+        varr, adj, nod_df = self._setup(t=400)
+        a = adj_corr(varr, adj, nod_df, freq=0.1).toarray()
+        b = adj_corr(
+            varr, adj, nod_df, freq=0.1,
+            positions=nod_df[["height", "width"]].values,
+        ).toarray()
+        assert np.allclose(a, b)
```
sneakers-the-rat: how does this test smoothing?

minian/test/test_cross_reg.py
```diff
@@ -8,6 +9,9 @@
 def test_cross_reg_notebook():
     os.makedirs("artifact", exist_ok=True)
     args = [
+        # Resolve jupyter via the active interpreter for Windows portability.
+        sys.executable,
+        "-m",
```
daharoni, line 14: Needed for Windows

minian/test/test_pipeline.py
```diff
@@ -34,12 +34,12 @@ def test_pipeline_notebook():
...
+    assert minian_ds.sizes["unit_id"] == pytest.approx(230, abs=10)
     motion_sum = minian_ds["motion"].sum("frame").values.astype(int)
     assert list(motion_sum) == pytest.approx([391, -252], abs=30)
     assert int(minian_ds["max_proj"].sum().compute()) == pytest.approx(1501702, rel=1e-2)
-    assert int(minian_ds["C"].sum().compute()) == pytest.approx(546290, rel=5e-2)
-    assert int(minian_ds["S"].sum().compute()) == pytest.approx(5065, rel=1e-1)
-    assert int(minian_ds["A"].sum().compute()) == pytest.approx(71468, rel=5e-2)
+    assert int(minian_ds["C"].sum().compute()) == pytest.approx(433137, rel=5e-2)
+    assert int(minian_ds["S"].sum().compute()) == pytest.approx(3669, rel=1e-1)
+    assert int(minian_ds["A"].sum().compute()) == pytest.approx(59311, rel=5e-2)
```
daharoni: With the bug fix for #302, these golden numbers have changed so updating them here.

minian/cnmf.py
```diff
@@ -1762,12 +1802,32 @@ def graph_optimize_corr(
...
+                [[G.nodes[n]["height"], G.nodes[n]["width"]] for n in nodes_sorted],
+                dtype=float,
+            )
+        except KeyError as e:
+            raise ValueError(
+                "graph_optimize_corr: positions=None fallback requires every "
+                "node to carry 'height' and 'width' attributes (missing key "
+                f"{e!s}). Either add these attrs when constructing G, or pass "
+                "positions= explicitly (e.g. footprint centroids for unit_merge)."
+            ) from e
```
sneakers-the-rat: sort of confusing to me why we would have two ways of specifying this. from what i can tell the old method didn't rely on height/width annotated on graph nodes, so it's not clear to me why we would add a new parameter `positions` as well as a fallback for it that isn't used anywhere except the tests in the same commit? why would we not just have one implementation that was either annotating the graph or passing positions?

minian/visualization.py
```diff
@@ -2191,3 +2191,148 @@ def visualize_motion(motion: xr.DataArray) -> hv.Layout | hv.NdOverlay:
...
             )
         )
+
+
+def visualize_spatial_partition(
+    max_proj: xr.DataArray,
+    positions: np.ndarray,
+    membership: np.ndarray,
+    adj: "scisps.spmatrix | None" = None,
+    n_frames: int | None = None,
```
daharoni: Possibly overkill but initially added it to visually verify our partition approach works as expected. This could be useful for very large or dense datasets. It also give a user some insight of what the code is doing which is one of the goals of minian.

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,422 @@
...
+        dims=("height", "width", "frame"),
+        coords={
+            "height": np.arange(height),
+            "width": np.arange(width),
+            "frame": np.arange(frames),
+        },
+    )
+
+
+class TestAdjCorrPositionPlumbing:
```
sneakers-the-rat: the relationship between this and `graph_optimize_corr` is not entirely clear to me, it seems like this is just a wrapper for that that constructs a graph, but it's not clear to me why it would be necessary - the names of the functions don't really help here. which of these is part of the public interface (which i guess is the pipeline notebook) and might be relied on downstream?

pipeline.ipynb
```diff
@@ -278,9 +278,12 @@
...
     "param_ks_refine = {\"sig\": 0.05}\n",
-    "param_seeds_merge = {\"thres_dist\": 10, \"thres_corr\": 0.8, \"noise_freq\": 0.06}\n",
-    "param_initialize = {\"thres_corr\": 0.8, \"wnd\": 10, \"noise_freq\": 0.06}\n",
-    "param_init_merge = {\"thres_corr\": 0.8}\n",
+    "# `chunk` is the target partition size handed to minian.cnmf.spatial_partition.\n",
+    "# Larger = fewer, bigger dask tasks; smaller = more, lighter tasks.\n",
+    "# Library default is 600.\n",
+    "param_seeds_merge = {\"thres_dist\": 10, \"thres_corr\": 0.8, \"noise_freq\": 0.06, \"chunk\": 600}\n",
+    "param_initialize = {\"thres_corr\": 0.8, \"wnd\": 10, \"noise_freq\": 0.06, \"chunk\": 600}\n",
+    "param_init_merge = {\"thres_corr\": 0.8, \"chunk\": 600}\n",
```
daharoni, line 286: the partition visualization works along side now having an adjustable chuck size. It generally won't need to be touched in most cases.

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,422 @@
...
+        ).toarray()
+        assert np.allclose(a, b)
+
+
+# ---------------------------------------------------------------------------
+# graph_optimize_corr: contract violations should fail loudly
+# ---------------------------------------------------------------------------
+
+
+class TestGraphOptimizeCorrContract:
```
sneakers-the-rat: so in neither of these do we actually test the functionality of the thing - that given some designed set of inputs, we get some expected set of outputs. we are only testing call conventions and other surface properties like shape and length matching. the thing we actually need for all these are tests that confirm the core operation does what it says it does across a range of possible inputs.

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,422 @@
...
+            graph_optimize_corr(
+                self._tiny_varr(), G, freq=None, positions=bad_positions
+            )
+
+    def test_rejects_missing_position_attrs(self):
+        G = self._build_graph(4, with_pos_attrs=False)
+        with pytest.raises(ValueError, match="positions=None fallback requires"):
+            graph_optimize_corr(self._tiny_varr(), G, freq=None, positions=None)
+
+    def test_membership_attrs_are_plain_int(self):
```
sneakers-the-rat: how would this not be true?

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,422 @@
...
+        # JSON-serializable; spatial_partition returns np.int64 and we cast.
+        G = self._build_graph(4)
+        membership = spatial_partition(
```
sneakers-the-rat: class says we are testing graph_optimize_corr but we don't call it here and instead call spatial_partition?

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,422 @@
...
+        parts = {n["id"]: n["part"] for n in json.loads(payload)["nodes"]}
+        assert all(isinstance(p, int) for p in parts.values())
+
+
+# ---------------------------------------------------------------------------
+# unit_merge centroid handling (NaN-safe + single fused compute)
+# ---------------------------------------------------------------------------
+
+
+class TestUnitMergeCentroidPathway:
```
sneakers-the-rat: don't understand this - `spatial_partition` does not call `unit_merge` and vice versa, but i see a comment in `unit_merge` about how some values are used by `spatial_partition` - when does this happen? what does it protect against? what is the NaN/FOV-centre fallback? why would we not be calling `unit_merge` in a test class that is named for it?

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,422 @@
...
+            assert hi - lo == 1
+
+
+# ---------------------------------------------------------------------------
+# partition_diagnostics
+# ---------------------------------------------------------------------------
+
+
+class TestPartitionDiagnostics:
+    """Arithmetic of the diagnostic helper, pinned on hand-built graphs."""
```
sneakers-the-rat: don't know what this means

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,422 @@
...
+
+    def _line_graph_membership(self, n: int, chunk: int):
+        # n nodes on a 1D line; stable argsort pairs consecutive nodes.
+        positions = np.column_stack([np.arange(n, dtype=float), np.zeros(n)])
+        membership = spatial_partition(positions, target_chunk=chunk)
+        adj = radius_neighbors_graph(positions, radius=1.5).astype(bool)
+        return positions, membership, adj
+
+    def test_sizes_and_n_parts_without_adj(self):
```
sneakers-the-rat: what does this test? what does it protect against?

sneakers-the-rat: similar question for rest of class

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,422 @@
...
+        assert diag["total_edges"] == 7
+        assert diag["edges_per_partition"].tolist() == [1, 1, 1, 1]
+        assert diag["cross_edges"] == 3
+        assert diag["cross_fraction"] == pytest.approx(3 / 7)
+
+    def test_symmetric_and_triangular_adj_give_same_result(self):
+        _, membership, adj = self._line_graph_membership(10, chunk=2)
+        from scipy.sparse import tril
+        diag_sym = partition_diagnostics(membership, adj=adj)
+        diag_tri = partition_diagnostics(membership, adj=tril(adj, k=-1))
```
sneakers-the-rat: what does this mean? when does it happen?

minian/cnmf.py
```diff
@@ -1857,28 +1928,132 @@ def adj_corr(
...
+    recurse(np.arange(n))
+    return membership
+
+
+def partition_diagnostics(
+    membership: np.ndarray,
+    adj: "scipy.sparse.spmatrix | None" = None,
+    n_frames: int | None = None,
+    bytes_per_sample: int = 4,
+) -> dict:
```
sneakers-the-rat, line 2067: the type here is not informative - what is in the dict? what do its fields mean? there is something in the docstring, but this isn't really good for modern python. this is especially informative for a diagnostic function - we have to know what's in the dict to use it.

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,422 @@
...
+        assert diag["mem_mb"].shape == (0,)
+
+
+# ---------------------------------------------------------------------------
+# chunk-kwarg plumbing
+# ---------------------------------------------------------------------------
+
+
+class TestChunkKwargPlumbing:
+    """Pipeline entry points expose `chunk` and forward it correctly."""
```
sneakers-the-rat: don't know what this means. are we just testing that a function forwards kwargs? why wouldn't we use a mock and test that the function is called with the given kwargs

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,422 @@
...
+    @pytest.mark.parametrize(
+        "func", [adj_corr, unit_merge, seeds_merge, initA]
+    )
+    def test_chunk_default_matches_constant(self, func):
+        sig = inspect.signature(func)
+        assert "chunk" in sig.parameters
+        assert sig.parameters["chunk"].default == DEFAULT_PARTITION_CHUNK
+
+    def test_adj_corr_returns_true_pearson(self):
+        # Regression for the vsub.T axis bug: adj_corr must match
```
sneakers-the-rat: i don't know what this means now, and nobody will in a month. if we are referring to a bug, we should refer to a raised issue or else give a description of the bug

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,422 @@
...
+        traces = np.stack(
+            [varr.isel(height=hs[k], width=ws[k]).values for k in range(n)]
+        )
+        for i, j in zip(*out.nonzero()):
+            truth = np.corrcoef(traces[i], traces[j])[0, 1]
+            assert out[i, j] == pytest.approx(truth, abs=1e-5)
+
+        small = adj_corr(varr, adj, nod_df, freq=None, chunk=4).toarray()
+        big = adj_corr(varr, adj, nod_df, freq=None, chunk=64).toarray()
+        assert np.allclose(small, big, atol=1e-5)
```
sneakers-the-rat: so we're asserting that the values are the same with different chunk sizes, how does this relate to some transposition bug? how does this correspond to the name of the function

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,422 @@
...
+        original = cnmf.graph_optimize_corr
+
+        def spy(*args, **kwargs):
+            seen_chunks.append(kwargs.get("chunk"))
+            return original(*args, **kwargs)
+
+        with patch.object(cnmf, "graph_optimize_corr", spy):
+            adj_corr(varr, adj, nod_df, freq=None, chunk=20)
+            adj_corr(varr, adj, nod_df, freq=None, chunk=5)
+        assert seen_chunks == [20, 5]
```
sneakers-the-rat: very strange assertion given the class and method name, i would assume we just construct a mock that carries the assertion that the kwargs are forwarded, this seems like a pretty indirect way to test, don't really understand it

minian/test/test_cross_reg.py
sneakers-the-rat, line 14: odd, if we are in a virtual environment we should receive the current environment which by definition contains a $PATH definition that puts the jupyter binary in the venv first. was this an experienced bug?

minian/test/test_pipeline.py
sneakers-the-rat: just want to note that as a reviewer this is why magic number tests are uninformative of code state - i have no idea if the old numbers are correct, i have no idea if the new numbers are correct, so code review devolves into "did you check out whatever this is and does it look good" and the answer is always "ya sure" without either party really being able to know for sure

minian/cnmf.py
```diff
@@ -45,6 +44,11 @@ def wrapper(fn: Callable) -> Callable:
         return wrapper
 
 
+#: Default target size for :func:`spatial_partition`. Forwarded by every
+#: ``chunk`` kwarg in this module and ``initialization.py``.
+DEFAULT_PARTITION_CHUNK = 600
```
sneakers-the-rat: why?

minian/cnmf.py
```diff
@@ -1469,7 +1474,35 @@ def unit_merge(
...
+            f"unit_merge: {int(bad.sum())} unit(s) have zero-mass footprints; "
+            "centroid positions replaced with FOV centre for partition "
+            "assignment (likely upstream update_spatial pruned to empty).",
+            RuntimeWarning,
+            stacklevel=2,
+        )
+        h_mid = float(A.coords["height"].mean().item())
+        w_mid = float(A.coords["width"].mean().item())
+        cen_h = np.where(bad, h_mid, cen_h)
+        cen_w = np.where(bad, w_mid, cen_w)
```
sneakers-the-rat: would be way cheaper to check if things are >0 rather than taking a sum, checking if two arrays are note finite, checking if any of the resulting array are true, then taking two more means and two more elementwise query and array constructors.

minian/cnmf.py
```diff
@@ -1469,7 +1474,35 @@ def unit_merge(
...
+    bad = ~np.isfinite(cen_h) | ~np.isfinite(cen_w)
+    if bad.any():
+        warnings.warn(
+            f"unit_merge: {int(bad.sum())} unit(s) have zero-mass footprints; "
+            "centroid positions replaced with FOV centre for partition "
+            "assignment (likely upstream update_spatial pruned to empty).",
+            RuntimeWarning,
+            stacklevel=2,
+        )
+        h_mid = float(A.coords["height"].mean().item())
```
sneakers-the-rat: this seems like we have allowed a bug in some other place and are compensating for it here, making another bug. when would it be correct to assume that the centroid of a zero-width object is the mean of all widths and heights? why wouldn't we just throw, both here and in the place that is generating the invalid input?

minian/cnmf.py
```diff
@@ -1789,9 +1849,10 @@ def construct_comput(edf, pxs):
             for d, dd in idx_dict.items()
         }
         vsub = varr.sel(**idx_arr).data
-        if len(idx_arr) > 1:  # vectorized indexing
-            vsub = vsub.T
-        else:
+        # `idx_corr` wants (N_items, T_frames). Vectorized indexing already
+        # produces that orientation; single-dim indexing needs a rechunk so
+        # the trace axis is contiguous for the numba kernel.
```
sneakers-the-rat: i don't *really* understand the bug here, partially because the prior code was unclear, but to document a prior bug inline in a comment the prior issue needs to be made clear. otherwise we are just adding mysteries. What was the prior bug? what did it do? why was it done that way, and what are we doing instead?

minian/cnmf.py
```diff
@@ -1857,28 +1928,132 @@ def adj_corr(
...
+    corr_df = graph_optimize_corr(
+        varr, G, freq, idx_dims=nod_df.columns, chunk=chunk, positions=positions
+    )
     return scipy.sparse.csr_matrix(
         (corr_df["corr"], (corr_df["source"], corr_df["target"])), shape=adj.shape
     )
 
 
-def adj_list(G: nx.Graph) -> list[np.ndarray]:
+def spatial_partition(
```
sneakers-the-rat, line 1935: so this is a new function. the types of tests that would be helpful here are a base case with explicitly specified positions that gives some obviously interpretable value that serves both as documentation and assurance that the core functionality is protected. then we want to have tests that systematically vary along the important elements of its inputs. e.g. `target_chunk` seems important, but in the tests `target_chunk` is seemingly randomly assigned to different values in different tests that seem like they have nothing to do with `target_chunk`. what does that do and how does its behavior vary along its values for the same data? does that do what we expect it to? what are the ways it might diverge from expected behavior, do we protect against those?

sneakers-the-rat, line 1935: needs more documentation about how the implementation works. since we are adopting a deep graph operation, we need to explain it. the LLM may have regurgitated some algorithm from its training data, but we need to understand and explain it.

minian/cnmf.py
sneakers-the-rat: this is particularly true since the body of the function is unreadable.

minian/visualization.py
```diff
@@ -2191,3 +2191,129 @@ def visualize_motion(motion: xr.DataArray) -> hv.Layout | hv.NdOverlay:
...
+        ``(N, N)`` adjacency (symmetric or triangular). Adds the intra-
+        partition edges histogram and the cross-partition fraction.
+    n_frames : int, optional
+        Length of the time axis. Adds a memory-per-partition histogram
+        (MiB, assuming float32).
+
+    Returns
+    -------
+    hvres : hv.Overlay or hv.Layout
+        ``hv.Overlay`` if neither ``adj`` nor ``n_frames`` is supplied;
```
sneakers-the-rat: why? what is the difference? how would they be used differently? when would adj and n_frames not be supplied, since this is a new function and it never called that way?

pipeline.ipynb
```diff
@@ -140,7 +140,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
```
sneakers-the-rat, line 143: versioning notebooks is a pain in the ass and we should have linter rules to avoid this kind of jitter in PRs

pipeline.ipynb
```diff
@@ -278,9 +278,12 @@
     "}\n",
     "param_pnr_refine = {\"noise_freq\": 0.06, \"thres\": 1}\n",
     "param_ks_refine = {\"sig\": 0.05}\n",
-    "param_seeds_merge = {\"thres_dist\": 10, \"thres_corr\": 0.8, \"noise_freq\": 0.06}\n",
-    "param_initialize = {\"thres_corr\": 0.8, \"wnd\": 10, \"noise_freq\": 0.06}\n",
-    "param_init_merge = {\"thres_corr\": 0.8}\n",
```
sneakers-the-rat, line 283: but the results are the same regardless of chunk size, right? that needs focused testing to protect against - there are two tests that seem to vary chunk size, but neither of them is something that "if it fails we know that the guarantee we are making to a user that the operation remains unchanged under chunk size by way of the comments has failed"

pipeline.ipynb
```diff
@@ -1888,6 +1893,42 @@
     "</div>"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Before running `seeds_merge`, preview the spatial partition it will use internally to chunk the correlation work. `spatial_partition` is deterministic in `(positions, chunk)`, so this renders the exact partition `seeds_merge` will use given the current `param_seeds_merge[\"chunk\"]`.\n",
```
sneakers-the-rat:
```suggestion
    "Before running `seeds_merge`, preview the spatial partition it will use internally to chunk the correlation work.",
```

not useful information to a human

pipeline.ipynb
```diff
@@ -1888,6 +1893,42 @@
...
+    "\n",
+    "What to look for:\n",
+    "- Balanced **nodes per partition** -- worker memory balance.\n",
+    "- Balanced **edges per partition** -- worker compute balance.\n",
+    "- Low **cross-partition edge fraction** (in the overlay title) -- shows up as `pixel recompute ratio` at runtime.\n",
```
sneakers-the-rat:
```suggestion
    "- Low **cross-partition edge fraction** (in the overlay title as `pixel recompute ratio`)\n",
```

we are human beings reading a document, it is runtime, we are not LLMs writing for a human being imagining what they will experience. we should be human beings writing for human beings.

pipeline.ipynb
```diff
@@ -1888,6 +1893,42 @@
...
+    "- **Memory per partition < per-worker budget** (`MINIAN_MEM_LIMIT / MINIAN_NWORKERS`).\n",
```
sneakers-the-rat: how would this be visible in the graph or be something we should look for?

pipeline.ipynb
```diff
@@ -1888,6 +1893,42 @@
...
+    "\n",
+    "Tune by editing `param_seeds_merge[\"chunk\"]` and re-running this cell. The same pattern applies to `param_initialize[\"chunk\"]` for `initA` and `param_init_merge[\"chunk\"]` for `unit_merge` further down."
```
sneakers-the-rat: and re-running the above cell where it is declared, right? what pattern? editing a value and re-running? or chunk size?

pipeline.ipynb
```diff
@@ -1888,6 +1893,42 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "hv.output(size=output_size)\n",
+    "from sklearn.neighbors import radius_neighbors_graph\n",
```
sneakers-the-rat: imports at the top

pipeline.ipynb
```diff
@@ -1888,6 +1893,42 @@
...
+    "\n",
+    "_seeds_in = seeds[seeds[\"mask_ks\"] & seeds[\"mask_pnr\"]].reset_index(drop=True)\n",
+    "_positions = _seeds_in[[\"height\", \"width\"]].values\n",
+    "_adj_preview = radius_neighbors_graph(\n",
+    "    _positions, radius=param_seeds_merge[\"thres_dist\"]\n",
+    ").astype(bool)\n",
+    "_membership = spatial_partition(_positions, target_chunk=param_seeds_merge[\"chunk\"])\n",
```
sneakers-the-rat: why are we declaring private variables in a notebook? the notebook is, reluctantly, an interface, so by the time when we are presenting something to someone in a notebook it should be the digested form where we only show them the inputs and outputs.

sneakers-the-rat (commented):
comments related to the random walk of LLM code. if we are cleaning things up, we can clean them up. if we are adding new things, we should test the new things and remove the things they replace. tests only really test argument parsing and don't make a net movement towards real coverage, miss important features of the new code. new code is introduced with fallbacks for conditions that should never happen or should be corrected for elsewhere.

minian/cnmf.py
daharoni: Yea this was already removed in a commit. 600 is used in a bunch of places so initially wanted to standardize where it was defined. But this is not how the rest of minian is written and I also made this parameter now tuneable through the partition visualization so it has been removed.

minian/cnmf.py
daharoni: Agree. This was a bug I stumbled across when starting to change things here. Removing the warning and just erroring out if mass <= 0 now.

minian/cnmf.py
daharoni: addressed in the commit for the change in the above comment. Now just using `(mass <= 0).any()`.

minian/cnmf.py
daharoni: seed_merge and initA does use this path but I agree it is confusing.

The change I am making is: `graph_optimize_corr` will require explicit positions (no more try/except), and 1adj_corr1 keeps a one-line derivation where `nod_df` already names the coords.

minian/cnmf.py
daharoni: added more detail in comment and also referenced #302

minian/cnmf.py
daharoni: A description of the approach is being added now. This was my solution to the problem that ended up, for obvious reasons, being a pretty common, named approach: k-d tree.

daharoni: Makes sense. I don't really know what ends up feeling like too much vs too little for testing things like this. I know the end goal is exactly what you describe above, just hard to know how to get there. 

That said, I think I added tests that do this now. Added tests that test across the input and output dimensions of the new method. They should both help describe and pin down the functionality of the new code.

minian/cnmf.py
daharoni: Yea I agree. I think we should end up moving to using pydantic or similar but nothing like that is used in the codebase yet so would prefer holding off. I added more detailed comments on the output dict for now.

minian/visualization.py
daharoni: I initially was only using parts of the visualization and this got left in. Removed now since, if visualizaing this part of the pipeline, all plots are needed and informative.

minian/test/test_cross_reg.py
daharoni: Yes, it is what caused early CI failures in Windows actions in the prior PR. I can double check that this is needed now but I am pretty sure it is.

minian/test/test_cnmf.py
daharoni: They are internal functions that solve different things: `seeds_merge` and `unit_merge` go through `adj_corr` but `initA` calls `graph_optimize_corr` directly because it already has a graph and it has sufficiently different needs than the merge methods.

minian/test/test_cnmf.py
daharoni: I don't know enough of testing approaches to be sure what you mean here but I tried moving some things into test fixtures.

minian/test/test_cnmf.py
daharoni: This isn't testing smoothing but rather testing adj_corr when it is doing smoothing internally. That said, it still isn't even testing that so will remove.

minian/test/test_cnmf.py
daharoni: roger that. I moved things around and should have the reasonable tests in `TestAdjCorr` now and added test_edge_correlations_match_corrcoef to make sure our edge corr coef calculation matches numpy's.

minian/test/test_cnmf.py
daharoni: removing

minian/cnmf.py
sneakers-the-rat: A typed dict does the trick! It is both a useful programmatic surface for static analysis and for documentation - with a typed dict here, when we get to docs we'll be able to autodoc it!

minian/test/test_cnmf.py
daharoni: I removed the FOV center fallback so this should now be resolved.

minian/test/test_cnmf.py
daharoni: Fixing to Verifies partition_diagnostics returns correct sizes, edge counts, and memory estimates for hand-built inputs with known expected values.

minian/test/test_cnmf.py
daharoni: Add comments for each saying what it protects against.

minian/test/test_cnmf.py
daharoni: answered in the comment I added. There are 2 ways to define the adj relationship between nodes/neurons. This just tests that they result in the same output

minian/test/test_cnmf.py
daharoni: agreed. rewrote using `patch.object(cnmf, "graph_optimize_corr", wraps=cnmf.graph_optimize_corr)`

minian/test/test_cnmf.py
daharoni: yea the test chunk default is 600  needs to be removed as it isn't doing anything useful. I think `test_chunk_kwarg_is_forwarded_to_graph_optimize_corr` is doing the test you suggested though.

minian/test/test_pipeline.py
daharoni: 100%.

pipeline.ipynb
daharoni: That would be great.

pipeline.ipynb
daharoni: We have `test_chunk_invariant` now to test this.

pipeline.ipynb
daharoni: We show all of this in the visualization of the spatial partitions. Percent edges shared across partitions, distribution of nodes in partitions, distribution of edges, and distribution of memory required per partition is all  plotted.

pipeline.ipynb
daharoni: Yea, this is currently handled exactly like all the other tunable parameters in minian. It just exposes the dict relevant to the section... you then can adjust the chunk value in it and rerun the visualization. Whatever you decide to stick with gets used in subsequent calculations.

In reality, this parameter was hardcoded as 600 previously and in 99% of cases would work just fine. We also could auto-optimize this value down the road but for now, I think exposing it visually in the same way other parameter are is good enough.

pipeline.ipynb
daharoni: updated the text a bit too to point users to the correct histogram.

pipeline.ipynb
daharoni: you are right. I moved these into 'visualize_seeds_merge_partition()' and update the notebook cell.

pipeline.ipynb
sneakers-the-rat: ah, well if the pattern that's being referred to is just that you can declare values in the notebook, i think we can omit that comment. hopefully it's obvious to people by line 1900 that they can declare params

minian/test/test_cnmf.py
sneakers-the-rat: ah, we can do some demos at some point, but fixtures are great for encapsulating reusable setup that separates setup code from test code, where all setup is injected into the test function by requesting it by fixture name in the test params

minian/cnmf.py
sneakers-the-rat: cool cool. basically i think we should have one way of specifying positions that is used ubiquitously - if that is embedding in node attributes in the nx graph, great, if that is passing a df/array of positions, also great. just one or the other so that we work towards having a small, clear set of data structures that are used throughout the package.

minian/test/test_cnmf.py
sneakers-the-rat: not necessary now, but this kind of structural documentation would be useful to have, the function names and docstrings don't really make it clear, and the lack of a clear public/private interface distinction (pre-existing in the package) makes it hard to know what is intended as an internal implementation detail and what is supposed to be exposed to a user (and thus documented to a level that a user could understand it)

minian/test/test_cross_reg.py
sneakers-the-rat: alright no prob, it's just in the tests anyway, so should be fine.

minian/cnmf.py
```diff
@@ -1937,22 +1926,69 @@ def spatial_partition(
...
     -------
     membership : np.ndarray
-        Integer partition label per row, dense in `[0, n_parts)`.
+        Integer partition label per row, dense in `[0, n_parts)`. Order
+        of labels is the order in which leaves are emitted by the
+        depth-first recursion (left-half before right-half at every
+        node), so consecutive labels correspond to spatially-adjacent
+        tiles.
+
+    Notes
```
sneakers-the-rat, line 1970: great, thanks for this additional explanation!

minian/cnmf.py
```diff
@@ -1469,7 +1469,33 @@ def unit_merge(
...
     print("computing temporal correlation")
     nod_df = pd.DataFrame({"unit_id": A.coords["unit_id"].values})
-    adj = adj_corr(C, A_inter, nod_df, freq=noise_freq)
+    # Footprint centroids for spatial_partition. Computed in a single
+    # dask.compute so the three reductions share one pass over A.
+    mass, mom_h, mom_w = da.compute(
+        A.sum(["height", "width"]).data,
+        (A * A.coords["height"]).sum(["height", "width"]).data,
+        (A * A.coords["width"]).sum(["height", "width"]).data,
+    )
```
sneakers-the-rat: I'm seeing that we already have a `calculate_centroids` and also a `visualization.centroid` function - can we use those/modify those/remove those so we have one way of doing things? seems odd to have this inlined here when we also have those separate functions. it also looks like this computation may be a bit different than `visualization.centroid`

minian/cnmf.py
```diff
@@ -1847,38 +1901,229 @@ def adj_corr(
     freq : float
         Cut-off frequency for the optional smoothing. If `None` then no
         smoothing will be done.
+    positions : np.ndarray, optional
+        `(N, 2)` array of 2D positions per node used for spatial chunking.
+        If `None`, derived from the ``height`` / ``width`` columns of
+        `nod_df`. Supply explicitly when the indexing columns aren't 2D
+        coordinates (`unit_merge` hands in footprint centroids).
```
sneakers-the-rat: still a bit squicky about having all these ways to specify the same thing - it looks like a `nod_df` constructed by `initialization.initA` has `width` and `height`, but then the `nod_df` that gets made in `unit_merge` only has `unit_id` in it and we pass positions separately. so the `nod_df` structure changes depending on who is calling us and the `height`/`width` fallback would fail.

it looks like `graph_optimize_corr` has three different sources of data depending on how we enter: 
- from `unit_merge`->`adj_corr` we have a position array derived from the `A` xarray. 
- from `seeds_merge`->`adj_corr` we don't pass positions, and derive that from a `seeds` dataframe.
- from `initA` we have a position array derived from node properties

it would be really great if we could have one way of representing the data and use it consistently, like it looks like within `graph_optimize_corr` we are required to have `height` and `width` defined as node attributes since we use that for `idx_dims`, so if that was the way to pass positions to `graph_optimize_corr`, and we just put `height` and `width` on the `nod_df` in `unit_merge` then we could avoid introducing a new data structure and make the existing ones more consistent.

minian/initialization.py
```diff
@@ -556,6 +561,7 @@ def seeds_merge(
     thres_dist=5,
     thres_corr=0.6,
     noise_freq: float | None = None,
+    chunk: int = 600,
```
sneakers-the-rat, line 564: add to docstring

minian/cnmf.py
```diff
@@ -1824,7 +1873,12 @@ def construct_comput(edf, pxs):
 
 
 def adj_corr(
-    varr: xr.DataArray, adj: np.ndarray, nod_df: pd.DataFrame, freq: float
+    varr: xr.DataArray,
+    adj: np.ndarray,
+    nod_df: pd.DataFrame,
+    freq: float,
+    positions: np.ndarray | None = None,
+    chunk: int = 600,
```
sneakers-the-rat, line 1885: add to docstring

minian/cnmf.py
```diff
@@ -1414,6 +1413,7 @@ def unit_merge(
     add_list: list[xr.DataArray] | None = None,
     thres_corr=0.9,
     noise_freq: float | None = None,
+    chunk: int = 600,
```
sneakers-the-rat, line 1416: add to docstring

minian/utilities.py
```diff
@@ -902,17 +902,29 @@ def custom_arr_optimize(
...
+    # override: it forces `fuse_keys` to be non-empty, and that interacts
+    # badly with TaskSpec's expression-layer optimisation. Return the
+    # graph unchanged on new dask; TaskSpec does its own optimisation
+    # internally and the legacy hooks here would just interfere.
+    # TODO: port the `tensordot` / `rechunk-merge` memory-throttle
+    # rewrites to `dask.annotate(resources=...)` at the call sites in
+    # cnmf.py / visualization.py so MEM throttling is restored on new
+    # dask -- those annotations are currently lost on this path.
+    if _DASK_HAS_TASK_SPEC:
+        return dsk
```
sneakers-the-rat: i wonder if we want to just set a minimum dask version `>=2025` and just drop this?

minian/utilities.py
```diff
@@ -921,22 +933,14 @@ def custom_arr_optimize(
...
-                    dsk[key_new] = val
-                    dsk[key] = key_new
+    effective_rewrite = rewrite_dict if rewrite_dict is not None else rename_dict
+    if effective_rewrite:
+        dsk_old = dsk.copy()
+        for key, val in dsk_old.items():
+            key_new = rewrite_key(key, effective_rewrite)
+            if key_new != key:
+                dsk[key_new] = val
+                dsk[key] = key_new
```
sneakers-the-rat: is this new version tested against `dask<2025`?

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,697 @@
...
+        membership = spatial_partition(grid_4x4, target_chunk=target_chunk)
+        counts = np.bincount(membership)
+        assert len(counts) == expected_n_parts
+        assert (counts == expected_part_size).all()
+
+    def test_max_part_size_never_exceeds_target_chunk(self):
+        # 200 points on a uniform grid, sweep target_chunk and assert
+        # the contract that no partition exceeds the requested size.
+        rng = np.random.RandomState(0)
+        positions = rng.uniform(0, 100, size=(200, 2))
```
sneakers-the-rat: ideally this would also randomly vary the number of points to counts that aren't divisible by 2, but probably fine.

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,697 @@
...
+        assert (unique == np.arange(len(unique))).all()
+
+
+class TestSpatialPartitionCompactness:
+    """The whole point of this function is to produce *spatially*
+    compact partitions -- a random label assignment would also satisfy
+    size constraints but cross many more edges in `graph_optimize_corr`.
+    """
+
+    def test_partition_diameter_is_bounded(self):
```
sneakers-the-rat, line 171: (this should be totally fine, i wonder what the odds are that this fails stochastically, like on the day we get a flaky test failure on this we should probably buy a lottery ticket)

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,697 @@
...
+        assert membership.tolist() == [0, 0, 0, 0]
+
+    def test_n_exceeds_target_chunk_by_one(self):
+        # 5 points, target_chunk=4: one split into 2 + 3.
+        positions = np.column_stack([np.arange(5, dtype=float), np.zeros(5)])
+        membership = spatial_partition(positions, target_chunk=4)
+        counts = np.bincount(membership)
+        assert sorted(counts.tolist()) == [2, 3]
+
+    def test_deterministic_across_repeated_calls(self):
```
sneakers-the-rat, line 238: not sure how this could fail since the algo is fully deterministic, but hey why not.

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,697 @@
...
+            # filt_fft_vec mutates in place; copy to keep `traces` clean for
+            # later inspection if this assertion fails.
+            traces_truth = filt_fft_vec(traces.copy(), freq, "low")
+        else:
+            traces_truth = traces
+        for i, j in zip(*out.nonzero()):
+            truth = np.corrcoef(traces_truth[i], traces_truth[j])[0, 1]
+            assert out[i, j] == pytest.approx(truth, abs=1e-5)
+
+    def test_fallback_matches_explicit_positions(self, adj_corr_setup):
```
sneakers-the-rat: a *little* bit spurious, since the fallback just directly undoes the operation here. if we wanted an integration test for matching results it would probably be to mock the function to catch the return value and call through the different callers listed in above comment

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,697 @@
...
+        for _, row in corr_df.iterrows():
+            i, j = int(row["source"]), int(row["target"])
+            truth = np.corrcoef(traces[i], traces[j])[0, 1]
+            assert row["corr"] == pytest.approx(truth, abs=1e-5)
+
+    def test_part_attrs_are_json_serializable(self, tiny_varr, line_graph_with_pos_attrs):
+        # spatial_partition returns np.int64; graph_optimize_corr must cast
+        # those to plain int before writing them onto graph nodes so the
+        # graph stays JSON-serializable. Without the cast, json.dumps
+        # raises ``TypeError: Object of type int64 is not JSON serializable``.
```
sneakers-the-rat: this is commented in `graph_optimize_corr` too - why are we concerned with json serializability here? is that because at some point we pass this data to a js plotting function or something?

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,697 @@
...
+        positions = np.zeros((5, 2), dtype=float)
+        membership = np.zeros(4, dtype=int)  # length mismatch
+        adj = scipy.sparse.csr_matrix((5, 5))
+        with pytest.raises(
+            ValueError,
+            match="positions has 5 rows but membership has 4",
+        ):
+            visualize_spatial_partition(
+                max_proj, positions, membership, adj=adj, n_frames=100
+            )
```
sneakers-the-rat: mv to `test_visualization`

sneakers-the-rat (commented):
looking gewd. the main thing i am still uneasy about is adding a new positions data structure when there are already 3 of them (node attributes, dataframe, positions on the xarray), it would be great if we could simplify that to derive positions from one of those, since the main caller of `spatial_partition` always has `varr` and `G` already and can extract `positions` from those. like it's fine for `spatial_partition` to receive a positions array, but the callers above it all derive it somewhat differently, and adding those new derivations pins us to them and makes it a bit harder to unify later.

Other than that, two minor things (consolidation of centroid computation, move the visualization test to test_visualization) and otherwise looks fine. the test cases for the spatial partitioning are pretty basic, but since it's just to optimize computation and doesn't affect the actual values of the correlation, nbd and we can add cases as we encounter bugs.

minian/cnmf.py
daharoni: Thanks for pointing this out. pymetis didn't need position information to partition the graph but our new method does. The wrinkle was that `unit_merge` didn't have the positions of the units calculated anywhere. 

That all said, I am going to:
* have `nod_df` always have hieght and width columns
* remove `positions` altogether
* move our zero-mass catch to after `centroid()` to make sure we don't drop units without knowing about it

minian/cnmf.py
daharoni: consolidating and more info is right below.

minian/test/test_cnmf.py
daharoni: moving

minian/utilities.py
daharoni: Yea, I think that is the way to go. There are enough changes with dask 2025 that I think we should just put a version floor for this and actually fully adjust the code for these. I'll do the minimum change here and there there should be a following (i'll open an issue) for some remaining dask changes.

minian/cnmf.py
sneakers-the-rat: Yes totally, the change from pymetis makes sense. I think it also makes sense for the spatial partitioning function to receive positions, that's a nice minimal pure function that takes in exactly what it needs and returns an array, just mostly the calling conventions above it propagating that positions array beyond the Function->proliferation of intermediate data forms that I think could be cleaned up. That multiple representations for the same thing is already inherent in the code, but I think we should probably work towards cleaning that up rather than adding new wrinkles.

minian/utilities.py
sneakers-the-rat: Cool cool. Yeah CalVer is all fun and games until it comes time to figure out whether "a year" actually corresponds to "a major version increment" and knowing what to pin.

sneakers-the-rat: https://github.com/dask/community/issues/372#issuecomment-4596547068

minian/test/test_cnmf.py
daharoni: lol. well this is here because there are multiple ways of solving the kd tree partition problem which can produce different but equally valid results. Because we compute this also for the optional visualization step, this test is just making sure we can't get different but equally valid outputs from the method. That said, it is currently fully deterministic so this is a trivial test but if we later adjust the method I just wanted to keep the output deterministic.

sneakers-the-rat: no prob, it runs super quick so why not have it!

minian/test/test_cnmf.py
```diff
@@ -0,0 +1,611 @@
...
+        # Uniform-grid points, sweeping both the point count -- including odd
+        # and prime counts not divisible by 2, where the median split can't
+        # halve evenly -- and target_chunk. No partition may exceed the
+        # requested size, for any combination.
+        rng = np.random.RandomState(0)
+        for n in (1, 2, 3, 7, 49, 199, 200, 201):
+            positions = rng.uniform(0, 100, size=(n, 2))
+            for tc in (1, 5, 17, 50, 200, 500):
+                membership = spatial_partition(positions, target_chunk=tc)
+                assert np.bincount(membership).max() <= tc, f"n={n} target_chunk={tc}"
```
sneakers-the-rat:
```suggestion
    @pytest.mark.parametrize("n", (1, 2, 3, 7, 49, 199, 200, 201))
    @pytest.mark.parametrize("chunk", (1, 5, 17, 50, 200, 500))
    def test_max_part_size_never_exceeds_target_chunk(self, n: int, chunk: int):
        # Uniform-grid points, sweeping both the point count -- including odd
        # and prime counts not divisible by 2, where the median split can't
        # halve evenly -- and target_chunk. No partition may exceed the
        # requested size, for any combination.
        rng = np.random.RandomState(0)
        positions = rng.uniform(0, 100, size=(n, 2))
        membership = spatial_partition(positions, target_chunk=chunk)
        assert np.bincount(membership).max() <= chunk
```

sorry if this comes across as being a busybody, mostly proposing just to show you that multiple parameterization is possible, this results in the product of the two conditions, and it makes it so if we were to run this with pytest-xdist the tests could be run in parallel, and also it gives the condition identifiers automatically so we don't need to remember to log those in the assertion

sneakers-the-rat (approved):
nice dog e. dogg i'm good with this

PR #305, opened 2026-05-30
Register minian on PyPI: publishing workflow + metadata

daharoni:
## Summary

- Adds a tag-triggered (`vX.Y.Z`) GitHub Actions workflow that builds an sdist + a single universal pure-Python wheel (`py3-none-any`) and publishes to PyPI via Trusted Publishing (OIDC) — no API token in the repo.
- Switches to **SCM-derived versioning** via `pdm-backend` (`source = "scm"`). The package version is read straight from the `v*` git tag at build time, eliminating drift across `pyproject.toml`, `[tool.commitizen]`, `minian/__init__.py`, and `minian/install.py`. Commitizen uses `version_provider = "scm"` and no longer writes version into source files. Mirrors the approach used in [`p2p-ld/numpydantic`](https://github.com/p2p-ld/numpydantic).
- Modernizes `pyproject.toml` for PyPI: SPDX license + `license-files`, Python 3.10–3.14 + topic trove classifiers, `Repository`/`Issues`/`Changelog` URLs, keywords, and a `[tool.pdm.build] excludes` rule that keeps `__pycache__` and numba AOT cache files (`*.nbi`/`*.nbc`) out of the artifacts.
- Refreshes `DeniseCaiLab/minian` → `miniscope/minian` URLs across the README, RTD config, install/contribute docs, and `minian-install`'s GitHub raw URL.
- Adds `RELEASING.md` with the one-time PyPI Trusted Publisher setup and the release flow (`pdm run cz bump` or a direct `git tag` for jumps).

## Before the first publish

One-time PyPI setup (per `RELEASING.md`): register `miniscope/minian` as a [pending publisher](https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/) for project name `minian`, workflow `publish.yml`, environment `pypi`. Then create a GitHub environment named `pypi` in this repo's settings.

The first PyPI version number is still TBD — could jump to `v2.0.0` or continue from the latest `v1.x` tag. Until a `v*` tag is pushed, SCM-derived dev builds resolve as `1.2.1.dev<n>+g<sha>` based on the latest existing tag.

## Test plan

- [x] `pdm build` locally produces a clean wheel + sdist, no `__pycache__` / `*.nbi` / `*.nbc` files
- [x] `twine check dist/*` passes on both artifacts
- [x] Wheel tag is `py3-none-any` (single universal wheel covers all OSes + Python 3.10–3.14)
- [x] SCM version derivation works (build produced `minian-1.2.2.dev90+g<sha>` from `v1.2.1`)
- [ ] After merge: `workflow_dispatch` from master builds artifacts successfully
- [ ] After PyPI trusted publisher is configured: real `v*` tag publishes to PyPI

Note on testing the workflow before merge: GitHub only registers `workflow_dispatch`-able workflows from the default branch, so `gh workflow run publish.yml` returns 404 on this feature branch. The workflow's `build` job runs the exact commands verified locally above. The test will be dispatched from `master` immediately after merge — it has no destructive side effects (only the `build` job runs on `workflow_dispatch`; `publish` is gated on `startsWith(github.ref, 'refs/tags/v')`).


RELEASING.md
```diff
@@ -0,0 +1,70 @@
...
+
+MiniAn is published to PyPI via GitHub Actions using
+[PyPI Trusted Publishing](https://docs.pypi.org/trusted-publishers/) (OIDC),
+so no API tokens are stored in this repo.
+
+The package version is derived from the **git tag** by `pdm-backend`'s SCM
+source. There is no `version = "..."` field to keep in sync — the version
+that ends up on PyPI is whatever you tag.
+
+## First PyPI release
```
sneakers-the-rat: I don't think we need the "first setup" part in the docs, but otherwise looks good

pyproject.toml
```diff
@@ -83,13 +122,13 @@ test = "pytest"
 
 [tool.commitizen]
 name = "cz_conventional_commits"
-version = "1.3.0"
+# `version_provider = "scm"` makes commitizen read the current version from
```
sneakers-the-rat, line 133: I'm still not exactly sure what committizen does, we can probably just remove it

sneakers-the-rat (approved):
Minor comments, otherwise lgtm

RELEASING.md
daharoni: Yea, it was just for me to remember what is left to do once we get pypi registered. Will remove before merge

daharoni:
`minian` in pypi is claimed but has never been published to. Maybe @phildong owns it... I have asked him but haven't heard back yet. I've also submitted a request to pypi for them to give us control. There isnt any other info on it.

sneakers-the-rat:
Perfect. Best outcome.

Needless to say @phildong you are still more than welcome to stay on as co-maintainer if you have time or inclincation to do so, not trying to take your project from you, just help it stay alive!

PR #307, opened 2026-06-01
feat: require FFmpeg/ffprobe before video I/O

thicclatka:
## Summary

MiniAn’s video load and export paths use FFmpeg via `ffmpeg-python`, but historically the code and docs **assumed** `ffmpeg` / `ffprobe` were already installed and on `PATH` — there was no upfront check and no explicit prerequisite in the install guide or README. Failures showed up later as opaque errors from `ffmpeg.probe`, pipe I/O, or `skvideo`.

This PR rectifies that:

- Adds `minian/ffmpeg_utils.py` with cached `ensure_ffmpeg()` and `FFmpegUnavailableError`, plus shared encode/load constants (`RawGray`, `H264`, `VideoExport`).
- Calls `ensure_ffmpeg()` before AVI/MKV ingest (`load_videos`, `load_avi_lazy`, `load_avi_ffmpeg`) and MP4 export (`write_video`, `write_vid_blk`, `concat_video_recursive`) so users get a clear message if binaries are missing.
- Documents FFmpeg as a prerequisite in `docs/source/start_guide/install.rst` and `README.md`.

Also includes a small `.gitignore` update to stop tracking `.DS_Store`.

## Context

Follow-up to the closed [#304](https://github.com/miniscope/minian/pull/304) discussion — incremental, reviewable slice targeting `modernize-minian`. Complements CI ffmpeg setup in [#303](https://github.com/miniscope/minian/pull/303) (runtime check vs install-in-CI).

## Test plan

- [x] With ffmpeg on PATH: `ensure_ffmpeg()` passes; `load_videos('./demo_movies', pattern='msCam1.avi')` → shape `(200, 480, 752)`; `pytest minian/test/test_pre_processing.py::test_can_load_videos` passed (~1s)
- [x] With ffmpeg removed from PATH (`env PATH=/usr/bin:/bin`): `ensure_ffmpeg` and `load_videos` raise `FFmpegUnavailableError` with install hint

See [comment](https://github.com/miniscope/minian/pull/307#issuecomment-4588721390) for exact commands and output.


minian/ffmpeg_utils.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Cached after the first successful check (or first failure message for stable
+    error reporting). Safe to call from hot paths (e.g. delayed workers): after
+    the first call it only branches on booleans.
+    """
+    global _success, _failure
+    if _success:
+        return
+    if _failure is not None:
+        raise FFmpegUnavailableError(_failure)
```
sneakers-the-rat:
```ipython
>>> %timeit shutil.which('ffmpeg')
51.6 μs ± 378 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

fast enough and used rarely enough that i think this could be a pure function

minian/ffmpeg_utils.py
```diff
@@ -0,0 +1,80 @@
...
+        try:
+            subprocess.run(
+                [exe, "-version"],
+                check=True,
+                capture_output=True,
+                timeout=30,
+            )
+        except (OSError, subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
+            _failure = f"{name!r} at {exe!r} failed to run (-version): {e}"
+            raise FFmpegUnavailableError(_failure) from e
```
sneakers-the-rat:
```suggestion
```

this, on the other hand, would be a bit slower. giving an informative error message in case people don't have ffmpeg installed is good, but i don't think we need to check for speculative error conditions. if `ffmpeg-python` exits with a nonzero exit code then it raises.

minian/ffmpeg_utils.py
```diff
@@ -0,0 +1,80 @@
...
+
+    MAX = 255
+    MIN = 0
+
+
+class VideoExport:
+    """Chunk sizing for video export helpers."""
+
+    STATS_REDUCE_FRAME_CHUNK_CAP = 32
+    CONCAT_LIST_CHUNK = 256
```
sneakers-the-rat: the current situation of having these defined inline (especially as mutable defaults) is not great, but if we are going to extract them out, the purpose should be to expose configuration to the user as something they could control in an intelligible way. this is something that's separable from raising an exception so it could get split into another PR or done here, but there are some problems with this as-is: 
- we are sort of defining keywords in an ad-hoc way with some special calling conventions within the body of the functions below, which makes an unnecessary layer of indirection over just having defaults that can be `**kwargs` spread into the ffmpeg-python functions
- the input format can't really be changed, it is necessarily fed via pipe and is always `format: rawvideo` and `pix_fmt: gray` by virtue of the data formats used within minian
- mixing different stages, filter kwargs and output kwargs on the H264 object, and we're also handling different kinds of input specially (some kwargs are declared on the class body, others are in a dict, and `PAD_FILTER` is a key not a value, `pix_fmt` is defined with two different keys)
- some of these are unused: `Uint8`, `VideoExport.STATS_REDUCE_FRAME_CHUNK_CAP`

`python-ffmpeg` doesn't define decent types, and the ffmpeg filter graph spec is a labyrinth, so we could define a TypedDict for the output kwargs but our type wouldn't be that informative, so this is one of those rare cases where a bare `dict` type annotation is actually appropriate.

If we wanted to change this, then we could do something like

```python
_DEFAULT_FFMPEG_OUTPUT = {
    "pix_fmt": "yuv420p",
    "r": 30,
    "...": "..."
}
```

and then down below in `write_video` we could do

```python
def write_vid_blk(
    # the rest of the kwargs...
    options: dict | None = None
):
    if options is None:
        options = _DEFAULT_FFMPEG_OUTPUT

    # ...
    process = (
        ffmpeg.input("pipe:", format="rawvideo", pix_fmt="gray", s=f"{w}x{h}")
        .filter("pad", int(np.ceil(w / 2) * 2), int(np.ceil(h / 2) * 2))
        .output(fname, **options)
        # ...
    )
    # ...
```

but as-is, not necessary for current goals of giving clear error message, and doesn't improve the - admittedly not optimal - magic number situation in the current code that much, and would need to be reworked into proper types.

minian/ffmpeg_utils.py
sneakers-the-rat, line 1: not a huge fan of the existence of a `utils` or `utilities` module in the first place, but as long as we have one, we might as well not have two - i think all we need from this module is the one function, so that can just go in `utilities`

minian/utilities.py
```diff
@@ -121,6 +123,7 @@ def load_videos(
 
     file_extension = os.path.splitext(vlist[0])[1]
     if file_extension in (".avi", ".mkv"):
+        ensure_ffmpeg()
```
sneakers-the-rat:
```suggestion
```

already called within `load_avi_lazy`

minian/utilities.py
```diff
@@ -266,9 +270,10 @@ def load_avi_ffmpeg(fname: str, h: int, w: int, f: int) -> np.ndarray:
     arr : np.ndarray
         The resulting array. Has shape (`f`, `h`, `w`).
     """
+    ensure_ffmpeg()
     out_bytes, err = (
         ffmpeg.input(fname)
-        .video.output("pipe:", format="rawvideo", pix_fmt="gray")
+        .video.output(RawGray.PIPE, format=RawGray.FORMAT, pix_fmt=RawGray.PIX_FMT)
```
sneakers-the-rat: these can't possibly have different values, no gain to constantizing them

minian/visualization.py
```diff
@@ -1219,7 +1221,7 @@ def write_video(
     vname: str | None = None,
     vpath: str | None = ".",
     norm=True,
-    options={"crf": "18", "preset": "ultrafast"},
+    options=H264.OUTPUT_OPTIONS,
```
sneakers-the-rat: still a mutable dict as default kwarg, unfortunately.

minian/visualization.py
```diff
@@ -1268,9 +1271,20 @@ def write_video(
...
-        ffmpeg.input("pipe:", format="rawvideo", pix_fmt="gray", s=f"{w}x{h}")
-        .filter("pad", int(np.ceil(w / 2) * 2), int(np.ceil(h / 2) * 2))
-        .output(fname, pix_fmt="yuv420p", vcodec="libx264", r=30, **options)
+        ffmpeg.input(
+            RawGray.PIPE,
+            format=RawGray.FORMAT,
+            pix_fmt=RawGray.PIX_FMT,
+            s=f"{w}x{h}",
+        )
+        .filter(H264.PAD_FILTER, int(np.ceil(w / 2) * 2), int(np.ceil(h / 2) * 2))
```
sneakers-the-rat: same as comment here: https://github.com/miniscope/minian/pull/307/changes#r3331741941

the input values can't possibly change, neither can the key value `PAD_FILTER`.

sneakers-the-rat (changes requested):
cool, thanks!

the check for ffmpeg makes sense, the current error message is just this:

```
FileNotFoundError: [Errno 2] No such file or directory: 'ffmpeg'
```

which for a new user would be confusing, so a link to the install page would be good. current PR seems to catch all the places ffmpeg would be invoked.

the rest of the changes i think are not ideal, constantizing these params in this way doesn't really add clarity since they are sort of haphazardly defined, rather than matching ffmpeg's call structure. So I think best path would be to just include the `ensure_ffmpeg` function and drop the rest of the changes. 

this is one reason why we recommend proposing changes in issues, so we can discuss changes first and avoid wasting your time - there are lots of improvements to be made in how I/O is handled in the package that are worth fixing, e.g. there are several different read and write methods that all work slightly differently (e.g. opencv is used sometimes, ffmpeg other times, there is i/o code in the visualization module, etc.) and that is worth fixing in a planned, coherent way, and the changes re: constantizing calling params just sort of bind us to the current structure at a distance rather than move us towards a better implementation.

so i think a nice, tidy, pure function that just calls `shutil.which` and raises an informative error would be great here. not much use in testing that since all it's doing is raising an error for lack of prereqs, so i don't think tests needed on that change.

sneakers-the-rat (approved):
hell ya, looks good :)

PR #312, opened 2026-06-01
ensure the lockfile is up to date with the current pyproject.toml

sneakers-the-rat:
add a check that ensures the lockfile is up to date w.r.t. the current state of pyproject.toml

note that this does not ensure that the lockfile is current *as in it has all the most recent dependencies locked* but rather that it reflects the most recent state of `pyproject.toml` (e.g. if a dep were to change in pyproject.toml without relocking, this would fail)


.github/workflows/lint.yml
```diff
@@ -25,13 +25,13 @@ jobs:
...
       - name: install python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.13"
-          cache: "pip"
+          python-version: "3.14"
       - name: install pdm
-        run: python -m pip install pdm
+        # exclude-newer in [tool.pdm.resolution] requires PDM >= 2.26.9; see pyproject.toml
+        run: python -m pip install "pdm>=2.26.9"
```
sneakers-the-rat, line 35: one can observe that the most recent version of pdm was installed without the explicit floor here - https://github.com/miniscope/minian/actions/runs/26782880616/job/78951497392#step:4:13

this would always be the case, pip always chooses the most recent version available for the platform. doesn't hurt to pin the floor, but it's also not needed.

daharoni: Sure. I see it was already doing this but I ran into this issue on my local computer so figured it doesn't hurt setting the floor here explicitly.

sneakers-the-rat: there is a weird chicken-and-egg problem with packaging frontends that makes this tricky, we have gone back and forth with the hatch devs in pyopensci about how to handle "we need to set a version floor for the packaging frontend tool, but versions of the frontend tool prior to the version that supports checking for the version floor of the packaging frontend tool will silently disobey the version floor anyway" - a strange omission in the python packaging specs that we're trying to get covered.

PR #313, opened 2026-06-01
Automatic weekly update of lockfile

sneakers-the-rat:
on monday at midnight, schedule a task to update the lockfile and open a PR with any changed deps. 

this does two things:
- avoids some PR noise from updating the lockfile during a PR (now the only lockfile changes that should happen in a PR are from changing a dependency in pyproject.toml)
- prevents us from getting out of date, where the versions we run tests against in CI become substantially different than those that someone would get from a `pip install`.

this is sort of a subtle and annoying thing about lockfiles: we want to use them so that local development environments are reproducible on CI, among other reasons, but using the lockfile means our installed package set will diverge from what happens with a normal install (whether from a pip install or if someone adds minian as a dependency, since solvers don't consider lockfiles of dependencies). so the lockfile needs to stay somewhat current, but it's a pain in the ass to always need to do that manually.

this uses an ~ official ~ blessed pdm action from the authors, so it should continue to work.


.github/workflows/lock-update.yml
```diff
@@ -0,0 +1,16 @@
+name: Update lockfile
+
+on:
+  schedule:
+    - cron: "0 0 * * 1"
```
daharoni, line 5: Oh! I never have looked into setting up a reoccurring job in github actions. That couldn't be easier.

sneakers-the-rat: yeah i have seen some people do some sicko shit automation with the free compute and cron tasks, it's an OP, underrated feature

PR #314, opened 2026-06-01
dep cooldowns in lockfile resolution!

sneakers-the-rat:
This is actually a feature that I added to PDM! (not that glamorous, the actual implementation was done, i just exposed it to pyproject.toml)

see here: https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

basically the tl;dr is that supply chain attacks that poison some upstream package are usually discovered within ~a week, so if we just exclude all packages newer than a week when resolving the lockfile we avoid a ~large proportion of supply chain attacks. since python still allows the old `setup.py` style of packaging, where to merely resolve the version of a package one must run arbitrary code, these supply chain attacks can be actually realized by both us and our contributors if they update the lockfile/install packages.


.github/workflows/testandcov.yml
```diff
@@ -60,7 +60,7 @@ jobs:
           fi
       - name: Install minian and test dependencies
         run: |
-          python -m pip install pdm
+          python -m pip install "pdm>=2.26.9"
```
daharoni, line 63: It looks like `exclude-newer` is only respected as of PDM 2.26.9. I found this out by regenerating the lock locally and it pulling in less than day old packages... turns out I needed to update my pdm.

sneakers-the-rat: yes :). i made sure to include that in the docs when i added it bc i knew that would happen eventually:

Screenshot 2026-06-01 at 8 05 36 PM

PR #316, opened 2026-06-01
Ci/real linting

sneakers-the-rat:
Fix: #311 

as #311 says - we didn't actually have linting/formatting enabled, just pyupgrade, so this is working us towards that.

there are _a lot_ of things to fix in rules that i think we definitely need to have enabled, most notably for annotations which can only be kinda sorta inferred automatically, and oh boy the code here has some twists and turns with all these local closures and callbacks and whatnot. 

it's good for us to be forced to take our medicine though! function annotations make the code *much* easier to work with, and who knows we might eventually get to a point where we can use mypy and numpydantic's type annotations to make the array passing a little less mysterious.

opening as a draft after just working through the `visualization` module, will come back to this another day, likely after we merge down the existing PRs.

current outstanding fixes to do are like this:

```
134     ANN001  missing-type-function-argument
 14     F405    undefined-local-with-import-star-usage
 13     NPY002  numpy-legacy-random
 12     ANN003  missing-type-kwargs
 12     ANN201  missing-return-type-undocumented-public-function
  6     ANN202  missing-return-type-private-function
  6     E722    bare-except
  4     ARG001  unused-function-argument
  4     S101    assert
  2     ARG002  unused-method-argument
  2     S110    try-except-pass
  1     B007    unused-loop-control-variable
  1     B904    raise-without-from-inside-except
  1     S113    request-without-timeout
  1     SIM108  if-else-block-instead-of-if-exp

```


minian/utilities.py
```diff
@@ -946,6 +909,12 @@ def custom_arr_optimize(
     # removing this now-vestigial optimizer + its arguments, is tracked
     # separately. The argument surface is kept for now so the call sites that
     # still pass these kwargs keep working until that follow-up lands.
+    if args or kwargs:
+        warnings.warn(
+            "Passing kwargs to custom_arr_optimized is deprecated and ignored",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
daharoni, line 897: This guard fires on every dask compute so a lot of deprecated warnings show up.

`custom_arr_optimize` is registered as dask's `array_optimize` (in `__init__.py` and the `da.config.set(array_optimize=...)` call sites), and dask always invokes it as `array_optimize(dsk, keys)`. So `*args` is always populated with `keys`, which means `if args or kwargs:` makes the `DeprecationWarning` fire on *every* array computation, including the bare registration that passes no deprecated kwargs at all (the message then misreports "Passing kwargs" when none were passed).

Keying the warning off `kwargs` only, the way the sibling `custom_delay_optimize` already does, keeps `*args` absorbing dask's positional `keys` silently and only warns when a call site actually passes the deprecated `keep_patterns` / `rename_dict` / ... kwargs. This also fixes the `custom_arr_optimized` typo in the message text.

```suggestion
    if kwargs:
        warnings.warn(
            "Passing kwargs to custom_arr_optimize is deprecated and ignored",
            DeprecationWarning,
            stacklevel=2,
        )
```

minian/motion_correction.py
```diff
@@ -378,8 +371,8 @@ def est_motion_chunk(
     prop_good = len(good_idxs) / len(good_fm)
     if prop_good < 0.9:
         warnings.warn(
-            f"only {prop_good} of the frames are good."
-            "Consider lowering your circularity threshold"
+            f"only {prop_good} of the frames are good.Consider lowering your circularity threshold",
```
daharoni: a missing space 

```suggestion
            f"only {prop_good} of the frames are good. Consider lowering your circularity threshold",
```

minian/utilities.py
sneakers-the-rat: I think since it's a passthrough that does nothing what we probably should do is just remove it entirely and not set it in the dask config. i set the deprecation warnings here to encourage us to resolve them in our internal code as well as to give a warning to anyone downstream that might have customized it in some way (i actually think that's impossible, since there's no surface to do that? but idk) that these no longer do anything. if we don't want to remove it for some reason, and keys is always passed, we could just add `keys` as an unused positional kwarg and still catch `*args` and `*kwargs` - to be informative, we need to emit the warning when a deprecated arg is used, whether its passed positionally or as a kwarg

sneakers-the-rat: here lets see if this passes https://github.com/miniscope/minian/pull/316/commits/91fa84edc8e06fcd8b300e3b7a2bb9187b6de046

PR #317, opened 2026-06-02
feat(packaging): ship notebooks in-package and fetch demo data from Zenodo

daharoni:
## What

Removes the ~700 MB of demo binaries (`demo_movies/`, `demo_data/`) from the repo and replaces them with on-demand fetching from Zenodo, and ships the tutorial notebooks inside the installed package.

## Changes

- **`minian/data/`** - new fetch/cache API (`fetch`, `dataset_path`, `datasets`) backed by `pooch`, a per-file registry with SHA256 checksums (`_registry.py`), and a `minian-data` CLI (`list`/`download`/`path`). Datasets are hosted on Zenodo (one deposit/DOI each) and verified on every access. `MINIAN_DATA_DIR` is an offline escape hatch.
- **`minian/notebooks/`** - the pipeline and cross-registration notebooks now live in-package as self-contained bundles (notebook + `assets/` + `README.md`). `minian-notebooks` CLI copies a bundle out to a working directory.
- **`minian/install.py`** - reduced to a thin back-compat shim delegating to the two new CLIs.
- **Tests** - notebook execution tests go through a shared `_notebook.py` helper; heavy end-to-end runs are marked `slow` (deselected by `pdm run test`, run in full by `test-all`/CI). Golden-value assertions unchanged.
- **CI/packaging** - a `prefetch-data` job caches datasets once per OS; `pdm-backend` ships notebooks in the wheel; `nbstripout` pre-commit hook + `notebooks-stripped` CI keep committed notebooks output-free.
- **`scripts/`** - `prefetch_data.py` (CI cache priming) and `zenodo_manifest.py` (maintainer staging tool).

## Notes

- Net working-tree change: **-698.7 MB** of binaries removed. Git history still contains the old blobs (a fresh clone stays large until history is rewritten separately).
- Both datasets are published on Zenodo and wired up in `_registry.py`.


docs/source/start_guide/install.rst
```diff
@@ -58,45 +58,41 @@ Note that if you install in this way you will have a local copy of MiniAn script
 Getting notebooks and demos
 ---------------------------
 
-The main features of Minian are exposed through `pipeline.ipynb` and `cross-registration.ipynb` `notebooks <https://jupyter.org/>`_.
-You can use the following links to get the latest version of the two notebooks:
+The main features of MiniAn are exposed through the `pipeline.ipynb` and `cross-registration.ipynb` `notebooks <https://jupyter.org/>`_.
+These notebooks ship **inside** the installed package, so once MiniAn is installed you already have them; no separate download is needed.
```
sneakers-the-rat: either this is a notice of change and goes in a changelog, or should be removed - people don't need to know about where things aren't and what they don't need to do

minian/install.py
```diff
@@ -1,75 +1,72 @@
...
-    else:
-        print(f"File {filename} not found with version {version}, skipping.")
 
+def notebook():
+    from minian.notebooks.cli import main as nb_main
+
+    print(
+        "Note: `minian-install --notebooks` is deprecated. "
+        "Use `minian-notebooks copy --all` (or `minian-notebooks list`)."
+    )
```
sneakers-the-rat: not exactly a deprecation warning, but good that we are preserving old calling paths with a warning!

to make this visible to people who might run this in a testing suite, we would also want to emit this as a proper DeprecationWarning (which are filtered by normal interactive use, like how normal stdout is in CI use, so in normal use someone would see one or the other but not both)

docs/source/start_guide/install.rst
```diff
@@ -58,45 +58,41 @@ Note that if you install in this way you will have a local copy of MiniAn script
...
-    
-    minian-install --notebooks
 
-Additionally, we also hosted some small demo data that works with the notebooks.
-Once you obtained these data, you should be able to run the two notebooks locally without modifying anything.
-Run the following script to get demo data:
+    minian-notebooks list              # show the available bundles
+    minian-notebooks copy pipeline     # -> ./minian-notebooks/pipeline/
+
+Each notebook fetches its demo recording automatically on first run (downloaded once, then cached and verified against a checksum), so there is no separate demo-install step.
```
sneakers-the-rat: same thing. people don't need to know about things they don't need to do.

docs/source/start_guide/install.rst
```diff
@@ -58,45 +58,41 @@ Note that if you install in this way you will have a local copy of MiniAn script
...
 
 .. code-block:: console
-    
-    minian-install --notebooks
 
-Additionally, we also hosted some small demo data that works with the notebooks.
-Once you obtained these data, you should be able to run the two notebooks locally without modifying anything.
-Run the following script to get demo data:
+    minian-notebooks list              # show the available bundles
+    minian-notebooks copy pipeline     # -> ./minian-notebooks/pipeline/
```
sneakers-the-rat: https://sphinx-argparse.readthedocs.io/en/latest/

docs/source/start_guide/install.rst
```diff
@@ -58,45 +58,41 @@ Note that if you install in this way you will have a local copy of MiniAn script
...
-    minian-install --demo
+    minian-data list                          # show datasets + sizes
+    minian-data download pipeline-demo --to .  # download into the current folder
+
+The demo datasets are hosted on Zenodo (each with its own citable DOI); see :mod:`minian.data` for details.
 
-The script can also help you get files from different branchs.
-See ``minian-install --help`` for more detail.
+.. note::
+    The older ``minian-install --notebooks`` / ``minian-install --demo`` commands still work as thin aliases for ``minian-notebooks`` and ``minian-data``.
```
sneakers-the-rat: deprecation warnings don't go in narrative docs - they can be put in API docs with a proper deprecation tag

docs/source/start_guide/install.rst
```diff
@@ -58,45 +58,41 @@ Note that if you install in this way you will have a local copy of MiniAn script
...
-    minian-install --notebooks
 
-Additionally, we also hosted some small demo data that works with the notebooks.
-Once you obtained these data, you should be able to run the two notebooks locally without modifying anything.
-Run the following script to get demo data:
+    minian-notebooks list              # show the available bundles
+    minian-notebooks copy pipeline     # -> ./minian-notebooks/pipeline/
+
+Each notebook fetches its demo recording automatically on first run (downloaded once, then cached and verified against a checksum), so there is no separate demo-install step.
+You can also fetch demo datasets yourself with the ``minian-data`` command:
```
sneakers-the-rat: add header sections for the different things

```
# notebooks

...

# data

...
```

minian/data/__init__.py
```diff
@@ -0,0 +1,143 @@
...
+
+def _make_pooch() -> pooch.Pooch:
+    """Build the pooch instance from the registry (one key per file)."""
+    registry, urls = {}, {}
+    for name, meta in _registry.DATASETS.items():
+        record = meta.get("zenodo_record")
+        for relpath, info in meta["files"].items():
+            key = f"{name}/{relpath}"
+            registry[key] = f"sha256:{info['sha256']}"
+            if record is not None:
```
sneakers-the-rat: not user data, hardcoded data, just get the key and throw a key error if record isn't present - we should consider it an error to have misspecified a dataset

minian/data/__init__.py
```diff
@@ -0,0 +1,143 @@
...
+                urls[key] = _registry.zenodo_url(record, info["zenodo"])
+    return pooch.create(
+        path=os.environ.get("MINIAN_CACHE_DIR", pooch.os_cache("minian")),
```
sneakers-the-rat: looks like pooch just has an `env` parameter that does this

minian/data/__init__.py
```diff
@@ -0,0 +1,143 @@
...
+        if not fpath.is_file():
+            raise FileNotFoundError(
+                f"MINIAN_DATA_DIR is set but {fpath} is missing for dataset {name!r}."
+            )
+        actual = _sha256(fpath)
+        if actual != info["sha256"]:
+            raise ValueError(
+                f"Checksum mismatch for {fpath}:\n  expected {info['sha256']}\n  got      {actual}"
+            )
+    return ddir
```
sneakers-the-rat: this is like the whole thing that pooch does. we should not duplicate the functionality of the thing we are using.

scripts/zenodo_manifest.py
```diff
@@ -0,0 +1,118 @@
...
+  Version:  v1
+  Related:  https://github.com/denisecailab/minian"""
+
+
+def sha256(path: Path) -> str:
+    h = hashlib.sha256()
+    with open(path, "rb") as f:
+        for chunk in iter(lambda: f.read(1 << 20), b""):
+            h.update(chunk)
+    return h.hexdigest()
```
sneakers-the-rat: duplicate of the function in data subpackage.

minian/data/cli.py
```diff
@@ -0,0 +1,76 @@
...
+        dest = Path(args.to)
+        dest.mkdir(parents=True, exist_ok=True)
+        for relpath in DATASETS[args.name]["files"]:
+            target = dest / relpath
+            target.parent.mkdir(parents=True, exist_ok=True)
+            shutil.copy2(src / relpath, target)
+        print(f"Copied {args.name} to {dest}")
+    else:
+        print(f"{args.name} ready at {src}")
+
```
sneakers-the-rat: why would we offer this method to create a duplicate copy of the data that is then never used?

minian/data/cli.py
```diff
@@ -0,0 +1,76 @@
...
+    parser = argparse.ArgumentParser(
+        prog="minian-data", description="Fetch MiniAn demo datasets."
+    )
+    sub = parser.add_subparsers(dest="command", required=True)
+
+    sub.add_parser("list", help="list available datasets").set_defaults(func=_cmd_list)
+
+    p_dl = sub.add_parser("download", help="download (and cache) a dataset")
+    p_dl.add_argument("name", help="dataset name")
+    p_dl.add_argument("--to", help="also copy the files into this directory")
```
sneakers-the-rat: it would typically be called `-o / --output`

minian/data/cli.py
sneakers-the-rat, line 1: i would typically organize these under a `cli/` subpackage rather than one cli module per feature, and then make a single `minian` cli entrypoint like `minian download ...`, `minian notebooks ...` rather than having several different cli entrypoints

minian/notebooks/cross_registration/cross-registration.ipynb
```diff
@@ -50,8 +49,14 @@
...
    "outputs": [],
    "source": [
-    "minian_path = \".\"\n",
-    "dpath = \"./demo_data/\"\n",
+    "import os\n",
+    "\n",
+    "from minian.data import fetch\n",
+    "\n",
+    "# `dpath` is the directory of session datasets to cross-register. By\n",
+    "# default we fetch the bundled two-session demo (cached, checksum-verified).\n",
```
sneakers-the-rat:
```suggestion
    "# default we fetch the bundled two-session demo.\n",
```

documentation for humans

minian/notebooks/pipeline/pipeline.ipynb
```diff
@@ -235,8 +234,11 @@
    "outputs": [],
    "source": [
     "# Set up Initial Basic Parameters#\n",
-    "minian_path = \".\"\n",
-    "dpath = \"./demo_movies/\"\n",
+    "# `dpath` is the folder containing the videos to process. By default we\n",
+    "# fetch the bundled demo recording (downloaded once, then cached and\n",
+    "# checksum-verified). To analyze your own data, set `dpath` to your own\n",
```
sneakers-the-rat:
```suggestion
    "# fetch the bundled demo recording.\n",
    "# To analyze your own data, set `dpath` to your own\n",
```

minian/notebooks/pipeline/README.md
sneakers-the-rat, line 1: why are all these done as READMEs within the code, where they will never be seen, rather than as docs pages?

minian/notebooks/cli.py
```diff
@@ -0,0 +1,148 @@
...
+def _iter_notebooks(node, prefix=""):
+    """Yield notebook paths (relative to the notebooks package root)."""
+    for entry in node.iterdir():
+        if entry.name in _SKIP:
+            continue
+        rel = f"{prefix}{entry.name}"
+        if entry.is_dir():
+            yield from _iter_notebooks(entry, prefix=rel + "/")
+        elif entry.name.endswith(".ipynb"):
+            yield rel
```
sneakers-the-rat:
```python
root = Path('some/path')
notebooks = [p.relative_to(root) for p in root.rglob("*.ipynb")]
```

minian/notebooks/cli.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def _bundles():
+    """Map bundle name (notebook-containing dir) -> sorted notebook relpaths."""
+    bundles: dict[str, list[str]] = {}
+    for rel in _iter_notebooks(_root()):
+        bundle = rel.rsplit("/", 1)[0] if "/" in rel else ""
+        bundles.setdefault(bundle, []).append(rel)
+    return {k: sorted(v) for k, v in sorted(bundles.items())}
```
sneakers-the-rat:
```python
root = Path("some/path")
parents = sorted([p for p in root.iterdir() if p.is_dir() and p.name not in _SKIP])
bundles = {p.name: sorted([child.relative_to(p) for child in p.rglob('*.ipynb')]) for p in parents}
```

don't use strings as paths, windows uses `'\'`. bundles are a single-layer phenomenon, don't recurse.

minian/notebooks/cli.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def _traverse(relpath):
+    node = _root()
+    for part in relpath.split("/"):
+        node = node.joinpath(part)
+    return node
```
sneakers-the-rat: what even

minian/notebooks/cli.py
```diff
@@ -0,0 +1,148 @@
...
+def _readme_summary(bundle):
+    try:
+        text = _traverse(f"{bundle}/README.md").read_text(encoding="utf-8")
+    except (FileNotFoundError, OSError):
+        return ""
+    for line in text.splitlines():
+        line = line.strip().lstrip("#").strip()
+        if line:
+            return line
+    return ""
```
sneakers-the-rat: not necessary. we don't need readmes for each pipeline. docs pages are better. for the sake of the CLI we just need descriptions. point people to the docs.

minian/notebooks/cli.py
```diff
@@ -0,0 +1,148 @@
...
+def _copy_tree(src, dest: Path):
+    if src.is_dir():
+        dest.mkdir(parents=True, exist_ok=True)
+        for child in src.iterdir():
+            if child.name in _SKIP:
+                continue
+            _copy_tree(child, dest / child.name)
+    else:
+        dest.parent.mkdir(parents=True, exist_ok=True)
+        dest.write_bytes(src.read_bytes())
```
sneakers-the-rat: what? `shutil.copytree`

minian/notebooks/cli.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def _cmd_list(args):
+    bundles = _bundles()
+    if not bundles:
+        print("No notebook bundles found.")
+        return
```
sneakers-the-rat: would never be the case

minian/notebooks/cli.py
```diff
@@ -0,0 +1,148 @@
...
+    width = max(len(b) for b in bundles)
+    for bundle, notebooks in bundles.items():
+        summary = _readme_summary(bundle)
+        print(f"{bundle:<{width}}  {summary}")
+        if len(notebooks) > 1:
+            for nb in notebooks:
+                print(f"{'':<{width}}    - {nb}")
```
sneakers-the-rat: if we had idk made some kind of "registry" class and then idk given it "methods" then we could maybe "reuse nearly-identical code that is duplicated between nearly identical modules"

minian/notebooks/cli.py
```diff
@@ -0,0 +1,148 @@
...
+        return _traverse(spec), Path(spec).name
+    # single notebook within a bundle
+    candidates = [spec, spec + ".ipynb"] if not spec.endswith(".ipynb") else [spec]
+    for rel in candidates:
+        node = _traverse(rel)
+        if node.is_file():
+            return node, Path(rel).name
+    raise SystemExit(
+        f"Unknown bundle or notebook: {spec!r}. Run `minian-notebooks list`."
+    )
```
sneakers-the-rat: this module is pretty spaghetti. this whole notion of bundles is unnecessary and could just be replaced with globbing. i have a directory of files. if someone gives me `thingname` i could just copy `thingname*`. and that's all i'd need to ever do if someone wanted `thingname` or `thingname/subname.ipynb` or anything else

minian/test/test_cross_reg.py
```diff
@@ -1,32 +1,22 @@
...
-    ]
-    subprocess.run(args, check=True)
-    assert os.path.exists("./demo_data/shiftds.nc")
-    assert os.path.exists("./demo_data/cents.pkl")
-    assert os.path.exists("./demo_data/mappings.pkl")
-    cents = pd.read_pickle("./demo_data/cents.pkl")
-    mappings = pd.read_pickle("./demo_data/mappings.pkl")
+    # The notebook fetches the two-session demo and writes its outputs back into
+    # that dataset directory, so resolve it here and read the outputs from there.
+    dpath = require_dataset("cross-reg-sessions")
```
sneakers-the-rat: textbook fixture use case

minian/test/test_cross_reg.py
```diff
@@ -1,32 +1,22 @@
...
+    execute_notebook("cross_registration/cross-registration.ipynb", "cross-registration")
+
+    assert (dpath / "shiftds.nc").exists()
+    assert (dpath / "cents.pkl").exists()
+    assert (dpath / "mappings.pkl").exists()
```
sneakers-the-rat, line 18: one big problem with these tests before is that they don't clear their output directories. this is why fixtures are good. on setup: ensure expected directory is cleared. on teardown: cleanup directory. make a tree of fixtures and then you can pluck from the tree of fixtures at any point. it's just dependency injection with cleanup and it's good.

minian/test/test_notebooks.py
```diff
@@ -0,0 +1,33 @@
...
+}
+
+_EXTRA = [nb for nb in discover_notebooks() if nb not in GOLDEN]
+
+
+def test_golden_notebooks_present():
+    """The golden-tested notebooks are actually shipped where tests expect."""
+    found = set(discover_notebooks())
+    missing = GOLDEN - found
+    assert not missing, f"expected bundled notebooks are missing: {sorted(missing)}"
```
sneakers-the-rat: we're testing our test function?

minian/test/test_notebooks.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@pytest.mark.skipif(not _EXTRA, reason="no non-golden notebooks to smoke-test yet")
+@pytest.mark.parametrize("relpath", _EXTRA)
+def test_extra_notebook_executes(relpath):
+    """Every non-golden bundled notebook at least runs end to end."""
+    execute_notebook(relpath, relpath.rsplit("/", 1)[-1].removesuffix(".ipynb"))
```
sneakers-the-rat: what is this concept of "golden notebooks" and why would we test some unknown future notebooks differently?

minian/test/test_pre_processing.py
```diff
@@ -19,9 +18,14 @@
...
 
-@pytest.fixture
+# Module-scoped: resolving the dataset hashes ~688 MB, so do it once and share
+# the (read-only) videos across the tests below.
+@pytest.fixture(scope="module")
 def varr():
-    return load_videos(dpath, **param_load_videos)
+    # Same demo recording as the pipeline notebook (the msCam .avi files);
+    # download/cache it, or skip if unavailable.
+    dpath = require_dataset("pipeline-demo")
```
sneakers-the-rat: great, yeah, see if we had a tree of fixtures then varr could request the dataset as a fixture, then from within test functions we can do the same thing, etc.

minian/install.py
```diff
@@ -1,75 +1,72 @@
...
+``minian-install`` used to download notebooks and demo data over HTTP from
+GitHub at install time. That is no longer how MiniAn ships:
+
+* Notebooks live **inside** the installed package; copy them out with the
+  ``minian-notebooks`` command.
+* Demo data is fetched on demand, cached, and checksum-verified by the
+  ``minian-data`` command.
+
+This command is kept as a thin, dependency-free alias for one or two releases
+so existing instructions keep working; please switch to the new CLIs.
```
sneakers-the-rat: a nice thing about grouping the cli altogether in a `cli` subpackage is that you don't need to do so much boilerplate for deprecations. just mark it `@deprecated` and remove it in the next version increment that you declare.

requirements/requirements-base.txt
sneakers-the-rat, line 1: let's just remove this

scripts/prefetch_data.py
sneakers-the-rat, line 1: why would we need a special script for this rather than just using our own CLI? `minian-data download --all`

scripts/prefetch_data.py
```diff
@@ -0,0 +1,54 @@
...
+import pooch
+
+_REGISTRY_PATH = Path(__file__).resolve().parent.parent / "minian" / "data" / "_registry.py"
+
+
+def _load_registry():
+    spec = importlib.util.spec_from_file_location("_minian_registry", _REGISTRY_PATH)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
```
sneakers-the-rat: code smells are at critical levels

scripts/prefetch_data.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def main():
+    reg = _load_registry()
+    registry, urls = {}, {}
+    for name, meta in reg.DATASETS.items():
+        record = meta.get("zenodo_record")
+        if record is None:
+            print(f"skip {name}: no zenodo_record yet")
+            continue
```
sneakers-the-rat: cli should fail if we have misspecified our registry.

pyproject.toml
```diff
@@ -93,32 +76,32 @@ build-backend = "pdm.backend"
...
     'pdm lock --with :all --python=">=3.12,<3.14" --append',
     'pdm lock --with :all --python=">=3.14" --append'
 ]
-test = "pytest"
+# Local default stays fast: the heavy end-to-end notebook runs are marked
+# `slow` and deselected here. `test-all` runs the full suite (unit + notebook)
+# in a single pass; CI uses it across the whole matrix because the pipeline
+# notebook is currently our main cross-version/cross-OS integration coverage.
+test = "pytest -m 'not slow'"
+test-all = "pytest"
```
sneakers-the-rat: this is definitely more fragile than the pytest-based version in showed here: https://github.com/miniscope/minian/pull/308#issuecomment-4594286826

we want to push config down towards the most local domain - specify in pyproject.toml rather than in gh workflows. specify in pytest config rather than pdm config, etc. so that if i or some LLM agent out there wants to run the tests and they just go `python -m pytest` rather than `pdm run test` they get the same behavior as they do when running the pdm script shorthand.

README.md
```diff
@@ -28,9 +21,14 @@ MiniAn is an analysis pipeline and visualization tool inspired by both [CaImAn](
...
-1. Optional install the demo movies: `minian-install --demo`
-1. Fire up jupyter: `jupyter notebook` and open the notebook "pipeline.ipynb"
+1. Copy the pipeline notebook into your working directory: `minian-notebooks copy pipeline`
+1. Fire up jupyter and open it: `jupyter notebook minian-notebooks/pipeline/pipeline.ipynb`
+
+The notebook downloads its demo recording automatically on first run (cached and
+checksum-verified), so there is no separate data-install step. Browse and manage
+the bundled notebooks with `minian-notebooks list` and the demo datasets with
+`minian-data list`. (The old `minian-install --notebooks` / `--demo` commands
+still work as thin aliases.)
```
sneakers-the-rat, line 39: make documentation for these public, link to the documentation.

sneakers-the-rat (commented):
it's like twice as large as it needs to be but it seems like it would mostly do the job. there are lots and lots of patterns that range from minorly disfavored to janky that are worth cleaning up. at their core these are basically two very similar features yet they duplicate many of the same functions and methods. All these readmes should move into the docs, the CLI should gain sphinx documentation if we're adding new clis, rm the stuff that redefines stdlib, etc.

minian/data/__init__.py
daharoni: cut the part pooch does

minian/test/test_cross_reg.py
daharoni: Done. We have `clean_dataset_outputs` now

minian/data/__init__.py
sneakers-the-rat: We are still duplicating pooch, pooch should provide the paths from the keys, and also pooch handles checksumming. 

when one does a fetch with pooch, the hashes are checked:
https://github.com/fatiando/pooch/blob/fe5fe70e4b969b49e0c6763b7fc0a89c44f20af4/pooch/core.py#L574
https://github.com/fatiando/pooch/blob/fe5fe70e4b969b49e0c6763b7fc0a89c44f20af4/pooch/core.py#L785

this could just be straight removed with no loss of function. this also lets us remove `_sha256`, and then in the case of partial downloads, pooch just silently re-downloads instead of throwing an error as we do here.

scripts/zenodo_manifest.py
```diff
@@ -0,0 +1,112 @@
...
+record id.
+"""
+
+import argparse
+import shutil
+import sys
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parent.parent
+sys.path.insert(0, str(REPO))
```
sneakers-the-rat: not sure why we need to insert the repo into sys.path - at the point when we are importing from the package, and thus its dependencies need to be installed, we might as well just directly import from it.

minian/data/__init__.py
daharoni: Right. make sense. working on removing this now.

minian/cli/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+__all__ = ["build_parser", "main", "data_main", "notebooks_main"]
+
+
+def build_parser() -> argparse.ArgumentParser:
+    """Construct the top-level ``minian`` parser (also used to render CLI docs)."""
+    parser = argparse.ArgumentParser(
+        prog="minian",
+        description="MiniAn command line tools: fetch demo data and copy out notebooks.",
+    )
+    sub = parser.add_subparsers(dest="group", required=True)
```
sneakers-the-rat: minor: sort of odd to have the subcommands called "positional arguments". i'd probably do `title='subcommands'`

pyproject.toml
```diff
@@ -74,7 +75,11 @@ Issues = "https://github.com/miniscope/minian/issues"
 Changelog = "https://github.com/miniscope/minian/blob/master/CHANGELOG.md"
 
 [project.scripts]
+# `minian` is the canonical CLI; the rest are thin deprecated aliases.
+minian = "minian.cli:main"
 minian-install = "minian.install:main"
+minian-notebooks = "minian.cli:notebooks_main"
+minian-data = "minian.cli:data_main"
```
sneakers-the-rat: `minian-notebooks` and `minian-data` are added by this PR and immediately deprecated? why not just not add them

minian/cli/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+def data_main(argv=None):
+    """Deprecated alias for ``minian data`` (the ``minian-data`` entrypoint)."""
+    _warn_alias("minian-data", "minian data")
+    main(["data", *(sys.argv[1:] if argv is None else argv)])
+
+
+def notebooks_main(argv=None):
+    """Deprecated alias for ``minian notebooks`` (the ``minian-notebooks`` entrypoint)."""
+    _warn_alias("minian-notebooks", "minian notebooks")
+    main(["notebooks", *(sys.argv[1:] if argv is None else argv)])
```
sneakers-the-rat: these commands are introduced in this PR - we shouldn't introduce new code that is immediately deprecated.

minian/data/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+Demo data lives outside the git repository (see :mod:`minian.data._registry`).
+:func:`fetch` downloads a named dataset once, caches it to the OS cache dir,
+verifies it against recorded SHA256 checksums, and returns the local directory::
+
+    from minian.data import fetch
+    dpath = fetch("pipeline-demo")   # pathlib.Path to a dir of msCam*.avi
+
+The cache location follows ``pooch.os_cache`` (e.g. ``~/.cache/minian`` on
+Linux, ``%LOCALAPPDATA%\\minian\\Cache`` on Windows) and can be overridden with
+the ``MINIAN_CACHE_DIR`` environment variable.
```
sneakers-the-rat, line 12: the `MINIAN_CACHE_DIR` doesn't show up in the docs anywhere since we aren't autodoc'ing modules and this isn't in the argparsers help texts. should be possible to know that one can do this from the docs or CLI helptext.

minian/data/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+
+Offline / air-gapped use: pooch verifies every file against its recorded
```
sneakers-the-rat, line 14: i love the idea of people out there running minian on an airgapped machined

minian/notebooks/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def notebook_root() -> Path:
+    """Filesystem directory holding the bundled notebooks."""
+    return Path(ir.files(__name__))
+
+
+def notebooks() -> dict[str, str]:
+    """Mapping of notebook name -> one-line description."""
+    return dict(NOTEBOOKS)
```
sneakers-the-rat: no-op

minian/test/_notebook.py
```diff
@@ -0,0 +1,62 @@
...
+    ``MINIAN_CACHE_DIR``) means the notebook's own ``fetch`` call hits the
+    cache. Skips the test if the dataset cannot be resolved.
+    """
+    try:
+        return dataset_path(name)
+    except OSError as exc:
+        # pooch raises requests/OS errors (a subclass of OSError) when it can't
+        # download and the file isn't already cached; skip rather than fail.
+        # An unknown dataset name (KeyError) still fails hard.
+        pytest.skip(f"demo dataset {name!r} unavailable: {exc}")
```
sneakers-the-rat: `dataset_path` calls fetch. we should fail not skip if the data can't be downloaded.

minian/test/conftest.py
```diff
@@ -10,3 +16,63 @@ def pytest_sessionstart(session):
...
+        for rel in outputs:  # setup: ensure the output slots are empty
+            _remove(dpath / rel)
+        cleaned.append((dpath, outputs))
+        return dpath
+
+    yield get_clean
+
+    for dpath, outputs in cleaned:  # teardown: leave only the input files
+        for rel in outputs:
+            _remove(dpath / rel)
```
sneakers-the-rat: close but not quite. this means that one needs to call `clean_dataset_outputs()` from within the test function for this to work. this is error prone.

this should work like:

```python
@pytest.fixture
def dataset(request):
    name = request.param
    dpath = require_dataset(name)

    yield dpath

    for output in DATASET_OUTPUTS[name]:
        # remove things..
```

and then from a function that wants to request the usage of some dataset

```python
@pytest.mark.parameterize("dataset", ["pipeline"], indirect=True)
def my_test(dataset):
    # ...
```

so that then the use of the dataset is tied to cleaning up after it.

minian/test/_notebook.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def execute_notebook(relpath, output):
```
sneakers-the-rat, line 20: this should assert the return code is 0

minian/test/test_pre_processing.py
```diff
@@ -19,9 +17,16 @@
...
+# Module-scoped: resolving the dataset hashes ~688 MB, so do it once and share
+# the (read-only) videos across the tests below. The dataset path is plucked
+# from the shared ``dataset`` fixture tree (defined in conftest) rather than
+# resolved directly here.
+@pytest.fixture(scope="module")
+def varr(dataset):
+    # Same demo recording as the pipeline notebook (the msCam .avi files);
+    # download/cache it, or skip if unavailable.
+    dpath = dataset("pipeline-demo")
+    return load_videos(str(dpath), **param_load_videos)
```
sneakers-the-rat, line 28: see above re: how to structure and parameterize fixtures

minian/test/test_pre_processing.py
```diff
@@ -19,9 +17,16 @@
 param_background_removal = {"method": "tophat", "wnd": 15}
 
 
-@pytest.fixture
-def varr():
-    return load_videos(dpath, **param_load_videos)
+# Module-scoped: resolving the dataset hashes ~688 MB, so do it once and share
+# the (read-only) videos across the tests below. The dataset path is plucked
+# from the shared ``dataset`` fixture tree (defined in conftest) rather than
+# resolved directly here.
```
sneakers-the-rat:
```suggestion
```

if we wanted to have the hashing happen only once, a `fetch_dataset` fixture could be session-scoped, and then a function-scoped `dataset` fixture could request it and handle cleanup

.gitignore
```diff
@@ -148,3 +148,12 @@ demo_data/shiftds.nc
 
 # MacOS
 .DS_Store
+
+# Zenodo upload staging (scripts/zenodo_manifest.py)
+zenodo_upload/
+
+# Local demo-data copy staged for Zenodo upload (scripts/zenodo_manifest.py --source)
+.minian_data/
```
sneakers-the-rat, line 156: probably also `minian-notebooks` right?

.github/workflows/lint.yml
```diff
@@ -35,3 +35,24 @@ jobs:
         run: python -m pip install "pdm>=2.26.9"
       - name: check lockfile
         run: pdm lock --check --with :all
+
+  notebooks-stripped:
```
sneakers-the-rat: we have a pre-commit config and a CI action - one can just run `pre-commit` directly so we only configure this once

sneakers-the-rat (commented):
definitely better. still some odd things here and there but p much there. main must fix here is not adding and immediately deprecating things, rest are mostly style suggestions for preventing us from going down a janky path

PR #320, opened 2026-06-03
perf(motion): ~12x faster motion estimation (dask dedup + cheaper kernel)

daharoni:
## Summary

Speeds up motion estimation (`minian/motion_correction.py`) by ~12-13x per core on realistic movie lengths, with no meaningful change in output. Two independent, compounding gains:

- **b1 - dask task-duplication fix.** `est_motion_part` exposed each chunk's `(template, shifts)` result as two separate dask graph nodes. After the dask >=2025 modernization, `custom_arr_optimize` became a pass-through, so the optimizer fused each leaf kernel into both consumers and emitted it under two keys, running `est_motion_chunk` **once per branch**. The recursion now keeps each chunk's result as a single delayed tuple and reduces via a `_est_motion_reduce` adapter, splitting only at the very end. (The old behavior was also subtly wrong: duplicated tasks re-ran the in-place-mutating kernel on shared views.)
- **b2 - cheaper per-frame kernel.** The rigid path now warps with `cv2.warpAffine` instead of `sitk.Resample`, and registers via a direct cross-correlation with parabolic sub-pixel refinement (`_xcorr_subpixel`) reusing per-frame `rfft2` results cached once per chunk. The non-rigid (BSpline/`mesh_size`) path is unchanged.

## Verified results

Baseline (pre-perf) vs this branch, demo movie 480x752, **single worker** (synchronous scheduler = total work):

| frames | blocks | old | new | speedup | dedup |
|---|---|---|---|---|---|
| 250  | 1 | 27.2s | 10.3s | 2.7x | 1.0x |
| 500  | 2 | 102.5s | 16.0s | 6.4x | 2.0x |
| 1000 | 4 | 367.5s | 31.6s | 11.6x | 4.0x |
| 2000 | 8 | 763.7s | 60.6s | **12.6x** | 4.0x |

Per-op micro-benchmarks (best of 5): transform `sitk.Resample` 4.1ms -> `cv2.warpAffine` 0.47ms (**8.8x**); registration skimage `phase_cross_correlation(upsample=100)` 65ms/pair -> cached parabolic xcorr 12ms/pair (**5.3x**).

The two axes decompose cleanly: the 1-block row is pure b2 (no duplication possible there); the `dedup` column is the b1 multiplier, which **plateaus at ~4x** (it scales with reduction-tree depth, ~log in block count, not linearly with length).

## Accuracy

- Registration shifts vs the old method: **maxΔ 0.043 px, meanΔ 0.009 px**.
- cv2 vs sitk warp: <1.7/255 max, ~0.06 mean intensity difference (interior), well below imaging noise.
- `est_motion_chunk` output is bit-for-bit identical to the pre-cleanup commit across 3D/4D x alt-error cases (the loop consolidation in this branch is behavior-preserving).

## Parallel scaling

These are single-worker (total-work) numbers; multi-worker speedup stacks on top. Measured on the new code (2000 frames / 8 leaf blocks, threaded scheduler, inner libs pinned to 1 thread): 2 workers 1.75x (88%), 4 workers 2.63x (66%), 8 workers 3.83x (48%). Efficiency is bounded by leaf-block count + the reduction tail, so it improves on longer recordings.

## Notes

- Base is `modernize-minian` (this branch was cut from it; it depends on the dask >=2025 modernization).
- Profiling/benchmark harnesses used to develop and verify this were intentionally kept local (untracked; `.gitignore` lists them) rather than shipped.


minian/motion_correction.py
```diff
@@ -208,62 +207,78 @@ def est_motion_part(
     if chunk_nfm is None:
         chunk_nfm = varr.chunksize[0]
     varr = varr.rechunk((chunk_nfm, None, None))
-    arr_opt = fct.partial(custom_arr_optimize, keep_patterns=["^est_motion_chunk"])
-    if kwargs.get("mesh_size", None):
+    has_mesh = bool(kwargs.get("mesh_size", None))
+    if has_mesh:
```
sneakers-the-rat: let's make this an official param rather than just in kwargs (along with the other params that are specified in the docstring but not declared?)

minian/motion_correction.py
```diff
@@ -208,62 +207,78 @@ def est_motion_part(
...
+        Aggregated template for the group.
+    motions : np.ndarray
+        Combined motion estimates for the group.
+
+    See Also
+    --------
+    est_motion_part
+    est_motion_chunk
+    """
+    tmps = np.stack([r[0] for r in res_grp], axis=0)
```
sneakers-the-rat, line 297: i have no idea if this would make a difference, and i sort of doubt it does, but purely out of curiosity i wonder if adding the stack to the dask processing graph improves things? i see the previous version did that, and i wonder if that helped or hurt.

minian/motion_correction.py
```diff
@@ -515,14 +608,18 @@ def est_motion_perframe(
     --------
     estimate_motion : for detailed explanation of parameters
     """
+    if mesh_size is None:
+        if src_fft is None:
+            src_fft = np.fft.rfft2(src)
+        if dst_fft is None:
+            dst_fft = np.fft.rfft2(dst)
+        return -_xcorr_subpixel(src_fft, dst_fft, src.shape)
```
sneakers-the-rat, line 646: is mesh size ever set? do we need the other leg still? or when is that useful vs. using this path without the mesh size?

minian/motion_correction.py
```diff
@@ -473,6 +513,52 @@ def est_motion_chunk(
...
+    shape : tuple[int, int]
+        Spatial shape `(height, width)` of the frames.
+
+    Returns
+    -------
+    sh : np.ndarray
+        Shift `(height, width)` of src relative to dst; `-sh` registers src onto
+        dst (the convention :func:`est_motion_perframe` returns).
+    """
+    cc = np.fft.irfft2(src_fft * np.conj(dst_fft), s=shape)
```
sneakers-the-rat: similar to the main optimization here (only do the forward ffts once), i wonder if we could also do this only once?

like above we get `src_fft` and `dst_fft` like 

```python
fft_mid = np.fft.rfft2(varr, axes=(-2, -1))
src_fft, dst_fft = fft_mid[i], fft_mid[didx]
```

I wonder if we could just do something like this

```python
fft_mid = np.fft.rfft2(varr, axes=(-2, -1))
fft_product = np.fft.irfft2(fft_mid * np.conj(fft_mid))
src_cc, dst_cc = fft_product[i], fft_product[didx]
```

that's strictly a naive suggestion without carefully considering the operation, so don't take that as a literal suggestion - just wondering if there is some way to also precompute this, as it is by far the most expensive part of this function

minian/motion_correction.py
```diff
@@ -473,6 +513,52 @@ def est_motion_chunk(
...
+    peak = np.unravel_index(np.argmax(cc), cc.shape)
+    c0 = cc[peak]
```
sneakers-the-rat: it would be _really great_ if we could avoid introducing new code that uses single letter or otherwise unreadable variable names - as the old adage goes, code is read orders of magnitude more often than it is written, so it's worth writing it out longform so it can be read later.

what does cc stand for? what does sh stand for? what are m and p in idx_m and idx_p? i can infer this from context and reading the docstring, but it would be easier if the variable names were informative

minian/motion_correction.py
```diff
@@ -473,6 +513,52 @@ def est_motion_chunk(
...
+    sh = np.empty(2)
+    for ax in range(2):
```
sneakers-the-rat: took me awhile to figure out what this was doing - a comment here would be nice

minian/motion_correction.py
```diff
@@ -492,7 +580,9 @@ def est_motion_perframe(
     dst : np.ndarray
         The destination frame of registration.
     upsample : int
-        Upsample factor.
+        Upsample factor. Only used for the non-rigid (`mesh_size`) path; the
+        rigid path estimates sub-pixel shifts by parabolic interpolation of the
+        cross-correlation peak (see :func:`_xcorr_subpixel`).
```
sneakers-the-rat: users will not be able to see `_xcorr_subpixel` because private functions are usually (correctly) excluded from the docs

minian/motion_correction.py
sneakers-the-rat, line 646: why flip sign on the outside of the function rather than in the function? is it always necessary to flip sign?

sneakers-the-rat (approved):
hell ya. so i don't really know about the method per se, so my comments are just about the code and pretty minor. i'll take your word for it on the correctness.

 so from what i understand from the description, we are basically carving out some interior part of `phase_cross_correlation` and using precomputed ffts? one question there is why that is done only on one leg of the perframe function? 

other than that seems good to me

minian/motion_correction.py
sneakers-the-rat: i guess "the other params" part was just github code folding making this function appear like it was another function, just this one - in general kwargs are for when you are forwarding something or not referring to their contents specifically, but if you are trying to get a specific value out of kwargs, just declare it

minian/motion_correction.py
daharoni: Nope ... well the non-rigid leg was here from before and generally isn't used by minian users. There is a use case for non-rigid motion correction but it actually isn't likely this implementation. And actually, now with rigid motion correction running quite quickly, the non-rigid approach I would want to implement (which builds off the rigid one) is feasible. I'll add a note in comments about this.

daharoni: it is just the convention previously in the code and needed downstream.

minian/motion_correction.py
daharoni: That was actually what was initially causing the major slowdown and resulted in dask doing some work twice.

minian/motion_correction.py
daharoni: I think `(fft_mid * np.conj(fft_mid))` would be just the autocorrelation of itself which always peaks at 0 so probably isn't a replacement.

minian/motion_correction.py
daharoni: 100%. I actually want to take a pass across the codebase and do this. for now I'll just improve things in this PR>

daharoni (commented):
Thanks for all of this. Addressed things and will merge after CI is green

minian/motion_correction.py
sneakers-the-rat: huh, super weird. you'd think making dask aware of the full task graph would make it better able to optimize over it, but that's just a naive assumption

minian/motion_correction.py
sneakers-the-rat: oh yeah not that exactly, basically that was just a placeholder for "whatever the actual operation would be," but that's just perf-hunt mode speaking, maybe there isnt a way to do a fully vectorized computation for this operation for all i know

PR #328, opened 2026-06-04
feat(simulation): physically-driven synthetic-data + teaching suite (minian.simulation)

daharoni:
**Draft / WIP.** Adds `minian.simulation`, a physically-driven synthetic calcium-imaging data generator plus a teaching scaffold. It produces ground-truth recordings used to test the pipeline (motion correction and CNMF recovery) and serves as a hands-on teaching entry point.

## What this adds
- Typed `Spec` surface and a Layer-2 physics/scene runtime (`spec.py`, `scene.py`, `steps/`).
- A composable step chain: tissue, cell, optics (incl. NA^2 light-collection efficiency), sensor, brain motion.
- `Recording`/`GroundTruth` types and a `simulate()` orchestrator with spec-hash to zarr local caching.
- A shared structural-metrics oracle (`metrics.py`) and a Cartesian spec sweep generator (`sweep.py`).
- Test suite: unit tests per module plus whole-pipeline structural self-consistency and per-stage motion-correction / CNMF footprint-recovery demos.

## Status
- Steps 2 to 13 implemented and committed (linear, one commit per step).
- Step 14 (the two training notebooks under `minian/notebooks/training/`) is in progress and intentionally left WIP.

## Base and dependencies
- Targets `v2-integration` (this is v2 work, not part of the v1.3 release on master).
- #303 (pymetis to k-d tree) is already in master, so Notebook 2's full CNMF chain runs on ordinary CI.
- Step 14 depends on the notebook + demo-data packaging work (issue #306, PR #317) landing on `v2-integration` first; the notebooks slot into that structure and inherit its discovery-test CI, so no bespoke notebook CI is added here.


minian/notebooks/training/01_anatomy/01_anatomy.ipynb
```diff
@@ -0,0 +1,469 @@
...
+   "id": "52b9261d",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import matplotlib.pyplot as plt\n",
+    "from matplotlib.colors import LogNorm\n",
+    "from scipy.ndimage import gaussian_filter\n",
+    "from ipywidgets import interact, FloatSlider, IntSlider\n",
+    "import mediapy\n",
```
sneakers-the-rat, line 55: missing dep, not sure that we need another dep for this? don't we already show videos somehow?

minian/simulation/spec.py
```diff
@@ -0,0 +1,834 @@
...
+# ---------------------------------------------------------------------------
+
+
+# Natural physical order of the pipeline. A step's domain is a class-level
+# attribute (not a serialized field); the Spec validator warns if the list
+# departs from this order.
+_DOMAIN_RANK: dict[str, int] = {"cell": 0, "tissue": 1, "motion": 2, "sensor": 3}
+
+
+class StepSpec(_Base):
```
sneakers-the-rat, line 455: why do step specs need to be different than the steps themselves? or, why are they in a different place? it seems like this might be for serialization/caching's sake, but it doesn't look like any of the steps bear any data on their own, so they would still have the same fields -> be just as serializable/hashable. just trying to get my head around 'what does a step consist of' and it seems like it's a step class, a stepspec, some additional fields in the ground truth, some additional fields in the scene, and maybe more? i don't see a declarative relationship between a stepspec and its step, just dynamically via the `build` method, so that would be nice to have. 

I think best case for a design here would be to have the steps be separable and composable, and to do that would need to have each step be somewhat self contained and not bleed into either one another or the containing classes like scene, but as-is there is a bit of crosstalk/undeclared dependencies like e.g. the `tissue.RenderStep` depends on `cell.CellOpticsStep` which depends on `cell.PlaceSomataStep`, but that wouldn't be obvious unless you read the code. 

the type annotations between a Step and its Spec need to use generics btw, because currently they are all receiving the base `Spec` class, and so all the attribute accesses are untyped, but if each step specifies its spec class with a typevar (or just has its spec as its attrs) then it would be possible to correctly resolve those types.

sneakers-the-rat: one way of handling dependencies would just be to declare them: like if `CellOpticsStep` depends on `PlaceSomataStep`, then a validator on the `steps` list could check that there is one of the necessary step instances prior to it in the list.

minian/simulation/spec.py
```diff
@@ -101,6 +101,21 @@ class Optics(_Base):
...
+    )
+
+    @field_validator("field_curvature_radius_um")
+    @classmethod
+    def _check_curvature(cls, v: float | None) -> float | None:
+        if v is not None and v <= 0:
+            raise ValueError(
+                f"field_curvature_radius_um ({v}) must be > 0, or None for a flat field."
+            )
+        return v
```
sneakers-the-rat, line 119: could just be

```python
from typing import Annotated as A
from annotated_types import Gt

class Optics(_Base):
    # ...
    field_curvature_radius_um: A[float, Gt(0)] | None
```

minian/notebooks/training/01_anatomy/01_anatomy.ipynb
daharoni: showing videos in notebooks in minian works really really poorly currently (slow and jumpy). I was seeing if mediapy happens to work better and it doesn't seem too.

PR #334, opened 2026-06-18
build(deps): use opencv-python-headless

daharoni:
## What

Switch minian's OpenCV dependency from `opencv-python` to `opencv-python-headless`.

## Why

A full-repo review confirmed minian makes **no cv2 GUI calls** anywhere (no `imshow`/`namedWindow`/`waitKey`/trackbars/window props, and no cv2 in the notebooks). Every cv2 symbol minian uses is in the core modules present in **both** builds:

- **imgproc**: `morphologyEx`, `GaussianBlur`, `medianBlur`, `bilateralFilter`, `matchTemplate`, `dilate`, `erode`, `findContours`, `arcLength`, `contourArea`, `adaptiveThreshold`, `warpAffine`, `connectedComponents`, `bitwise_and`, `cvtColor`
- **videoio**: `VideoCapture` + `CAP_PROP_*` (headless bundles ffmpeg)
- **core**: `setNumThreads` / `getNumThreads` (bench scripts)

`opencv-python-headless` differs from the full build *only* by omitting GUI windowing and the `libGL`/X11/Qt system-lib requirement — none of which minian touches. So this is a packaging change with no functional impact.

Benefits:
- Drops the `libGL`/X11/Qt system-lib requirement → lighter, more robust CI and container installs.
- Lets minian share a single `cv2` build with downstream consumers that use headless. In particular [minisim](https://github.com/miniscope/minisim) (slated as a test dependency for simulated ground-truth data) depends on `opencv-python-headless`; without this change the two would co-install two conflicting `cv2` builds in one env.

## Lockfile churn

The `pdm.lock` re-resolve also advances **ipykernel 7.2.0 → 7.3.0** (which depends on `nest-asyncio2` upstream) plus minor `pyviz-comms`/`mistune` bumps. These are **forced by keeping the Python 3.14 lock target** (only ipykernel ≥7.3.0 supports 3.14), not by the opencv change — they'd arrive on the next re-lock regardless. Nothing brand-new slips through; the repo's `exclude-newer = 7d` guard still applies.

----
📚 Documentation preview 📚: https://minian--334.org.readthedocs.build/en/334/

### DIFF
--- a/pdm.lock
+++ b/pdm.lock
@@ -5,7 +5,7 @@
 groups = ["default", "doc", "numba", "test"]
 strategy = ["inherit_metadata"]
 lock_version = "4.5.0"
-content_hash = "sha256:716dc097be33f73ef7df4d8102d1466b06d2d410cbc91918a6bdff2d2102de32"
+content_hash = "sha256:90c4667bf369a6ceb754b38f1bf6cb5ee5f4306e018d5668b5f2dce0486dc5a6"
 
 [[metadata.targets]]
 requires_python = ">=3.10,<3.12"
@@ -1559,7 +1559,7 @@ files = [
 
 [[package]]
 name = "ipykernel"
-version = "7.2.0"
+version = "7.3.0"
 requires_python = ">=3.10"
 summary = "IPython Kernel for Jupyter"
 groups = ["default"]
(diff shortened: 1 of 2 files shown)


sneakers-the-rat (approved):
i am glad the LLM was able to grep for `cv2` and find no gui calls lol. no explanation needed

PR #335, opened 2026-06-18
test(motion): unit-test motion correction against minisim ground truth

daharoni:
> **Draft / stacked on #334.** Built on `deps/opencv-headless` so the test env uses a single (headless) `cv2`. The diff will shrink to just the test changes once #334 merges into `master` and this rebases.
>
> **Lock pending.** `pdm.lock` does **not** yet include `minisim`: every minisim release is <7 days old, so the repo's `exclude-newer = 7d` guard refuses to resolve it. Regenerating the lock (and marking this ready) is deferred until minisim 1.0.0 ages past the 7-day window (~2026-06-25). Until then CI's `pdm lock --check` will fail by design.

## What

The first stage-level unit tests that score minian against synthetic recordings carrying **exact ground truth** ([minisim](https://github.com/miniscope/minisim)), so motion-correction coverage no longer rests solely on the slow end-to-end pipeline notebook.

### `minian/test/_simulated.py` (reusable infrastructure)
A small 256×256 scope (~1.66 µm/px, ~80–90 detectable CA1 cells for rich registration texture), a cached `simulate_recording()` factory, and `as_movie()` to adapt a recording's `observed` counts into the dask-backed `(frame, height, width)` `DataArray` the pipeline consumes. Intended as the basis for **future** stage tests (preprocessing, init, CNMF), not just motion.

### `minian/test/test_motion_correction.py`
- **Kernel tier** (numpy-only, deterministic, microseconds): integer + sub-pixel shift recovery via `est_motion_perframe`, FFT-cache fast-path equivalence, `transform_perframe` round-trip and translation direction.
- **Ground-truth tier** (minisim): `estimate_motion` recovers a random-walk trajectory **and** an exact prescribed `trajectory_um` to **<1 px aligned RMSE** (measured ~0.5–0.7 px); a static movie estimates near-zero; `apply_transform` sharpens the center-cropped temporal mean.

Scored with `minisim.shift_rmse(..., correction=True, align=True)`: minian estimates the *correction* (negation of `GroundTruth.shifts`), and `align` absorbs the constant template-origin offset.

### Wiring
- `conftest.py`: session-scoped `simulate_recording` fixture (lazy minisim import) as the shared entry point.
- `pyproject.toml`: `minisim>=1.0,<2` added to the `[test]` extra.

All 13 tests pass locally in ~3 s; ruff check + format clean. The slow notebook golden stays as the integration anchor.

----
📚 Documentation preview 📚: https://minian--335.org.readthedocs.build/en/335/


minian/test/test_motion_correction.py
```diff
@@ -0,0 +1,146 @@
...
+)
+from ._simulated import DURATION_S, FPS, PIXEL_SIZE_UM, as_movie
+
+# --- kernel: registration + warp primitives (no simulator) -----------------
+
+
+def _smooth_frame(h: int = 96, w: int = 96) -> np.ndarray:
+    """A smooth, non-periodic-enough texture for sub-pixel registration."""
+    yy, xx = np.mgrid[0:h, 0:w]
+    return (np.sin(yy / 7.0) * np.cos(xx / 5.0) + 1.5).astype(np.float32)
```
sneakers-the-rat, line 37: why not just use the simulation framework to generate a single video sessionwide to use for generic purposes as a fixture and then use it in all these tests

daharoni: `_smooth_frame` is just a fake clean frame that we shift by a known amount to quickly check the basic frame-to-frame alignment. 

For the actual `estimate_motion` / `apply_transform` tests we do use the simulation. Each of those tests makes up its own little simulation with its own motion (random walk, set trajectory, or no motion). A shared video couldn't cover multiple cases. But we basically can constructure variations of the same video structure with very little overhead here.

PR #340, opened 2026-06-25
Fix visualize_seeds crash on HoloViews ≥1.16 (size_index/color_index removed)

daharoni:
## Problem

`visualize_seeds()` crashes on any install that resolves HoloViews >= 1.16:

```
ValueError: Unexpected option 'size_index' for Points type across all extensions. No similar options found.
```

`size_index` and `color_index` were removed from `hv.Points` in HoloViews 1.16. There is no upper bound on HoloViews in `requirements.txt`, so a fresh `pip install` pulls the latest (1.23.0) and `visualize_seeds()` fails immediately. Reported against Minian 2.0.0, HoloViews 1.23.0, Python 3.12.

## Fix

In `visualize_seeds` (`minian/visualization.py`), replace the removed `*_index` style options with their modern dimension-mapping equivalents — **preserving the original rendering**, not just silencing the error:

- `color_index=mask` -> `color = mask` (maps the mask column directly; existing `pt_cmap` still colors filtered-out seeds red).
- `size_index="seeds"` -> `size = (dim("seeds") * 6) ** 0.5`. The old path, with no explicit `size`, rendered each point with radius `sqrt(base_size * seeds)` where `base_size = size**2 = (sqrt(6))**2 = 6` (area scaling, `scaling_factor=1`). This expression reproduces that exactly, so the seed-count size encoding (a seed found as a local max in more chunks shows as a bigger dot) is retained.

## Verification

- Reproduced the exact `ValueError` with the original code on HoloViews 1.23.0 + Minian 2.0.0.
- Confirmed the patched function renders cleanly (`hv.render`, both `mask=None` and `mask="mask_pnr"` paths).
- Confirmed the rendered bokeh glyph sizes are **bit-for-bit identical** to the old `size_index` path across `seeds in {1,2,5,10,25,50,100}` on HoloViews 1.22.1 (which still implements `size_index`), and that the new `size=` expression renders on 1.23.0.

Reported via miniscope/workshop-processing-and-analysis#1.

Generated with Claude Code

### DIFF
--- a/minian/visualization.py
+++ b/minian/visualization.py
@@ -1569,15 +1569,16 @@ def visualize_seeds(
     opts_pts = {
         "frame_width": 600,
         "aspect": asp,
-        "size": 8,
+        # HoloViews >=1.16 removed size_index/color_index; map the dimensions
+        # directly instead. size_index="seeds" rendered each point with radius
+        # sqrt(base_size * seeds) where base_size = size**2 = (sqrt(6))**2 = 6
+        # (area scaling, scaling_factor=1), so reproduce it exactly here.
+        "size": (hv.dim("seeds") * 6) ** 0.5,
         "tools": ["hover"],
         "fill_alpha": 0.8,
         "line_alpha": 0,
         "cmap": pt_cmap,
     }
-    # HoloViews >=1.16 removed the size_index/color_index options; map the color
-    # dimension directly instead (the mask column, via pt_cmap) and use a fixed
-    # point size.
     if mask:
         vdims = ["seeds", mask]
         opts_pts["color"] = mask
--- a/minian/visualization.py
+++ b/minian/visualization.py
@@ -1569,15 +1569,18 @@ def visualize_seeds(
     opts_pts = {
         "frame_width": 600,
         "aspect": asp,
-        "size_index": "seeds",
-        "color_index": mask,
+        "size": 8,
         "tools": ["hover"],
         "fill_alpha": 0.8,
         "line_alpha": 0,
         "cmap": pt_cmap,
     }
+    # HoloViews >=1.16 removed the size_index/color_index options; map the color
+    # dimension directly instead (the mask column, via pt_cmap) and use a fixed
+    # point size.
     if mask:
         vdims = ["seeds", mask]
+        opts_pts["color"] = mask
     else:
         vdims = ["seeds"]
         opts_pts["color"] = "white"


minian/visualization.py
```diff
@@ -1569,15 +1569,19 @@ def visualize_seeds(
     opts_pts = {
         "frame_width": 600,
         "aspect": asp,
-        "size_index": "seeds",
-        "color_index": mask,
+        # HoloViews >=1.16 removed size_index/color_index; map the dimensions
+        # directly instead. size_index="seeds" rendered each point with radius
+        # sqrt(base_size * seeds) where base_size = size**2 = (sqrt(6))**2 = 6
+        # (area scaling, scaling_factor=1), so reproduce it exactly here.
```
sneakers-the-rat, line 1575:
```suggestion
```

who is this comment for? why is it important that we document that the code does the same thing as some old code that is no longer in the repo?

sneakers-the-rat (approved):
weird how claude misidentified the version of removal in the original issue and then that misidentification is repeated here, but then it contradicts itself with the correct information in the same issue:

the options were deprecated in [1.11](https://holoviews.org/releases.html#version-1-11-0) and removed in [1.23](https://holoviews.org/releases.html#version-1-23). They aren't even options on the `Points` class, they are options on the containing plot object, which can be seen in the PR that removes them - https://github.com/holoviz/holoviews/pull/6819/changes

It says:

> size_index and color_index were removed from hv.Points in HoloViews 1.16.

but also:

> sizes are bit-for-bit identical to the old size_index path across seeds in {1,2,5,10,25,50,100} on HoloViews 1.22.1 (which still implements size_index), and that the new size= expression renders on 1.23.0.

also strange is blaming the lack of an upper bound in `requirements.txt` when there is no `requirements.txt` - 

> There is no upper bound on HoloViews in requirements.txt

I didn't dig into the code to evaluate the mapping expression, strange that it used to apply some exponential transform without being asked to do so, but if the output was actually validated as identical, i guess good enough for me

minian/visualization.py
daharoni: This comment is for me. This is the exact mapping of number of seeds to point radius that minian previously relied on when it was in HoloViews. 

This mapping is somewhat arbitrary so the comment here is to explain where it came from and why it is currently here. Without it, down the line it might be confusing as to explaining the somewhat arbitrary mapping if we want to change it.

PR #341, opened 2026-06-25
Fix save_minian crash on xarray >=2024 (strip stale chunk encoding)

daharoni:
## Problem

On xarray >= 2024.x, `save_minian()` raises during `ds.to_zarr()`:

```
ValueError: Specified Zarr chunks encoding['chunks']=(1, 2000) for variable named 'C_lp'
would overlap multiple Dask chunks.
```

This fires when the array being saved carries a **stale** zarr `encoding['chunks']` — inherited because it was read back from a previous `save_minian` via `xr.open_zarr` and then rechunked in memory (e.g. by `update_spatial` / `lowpass_temporal`). The stored chunk shape no longer tiles the array's current dask layout, so xarray refuses the write. In the no-deconvolution pipeline notebook this hits `C_lp = save_minian(C_chk.rename("C_lp"), ...)`, but it is a general `save_minian` issue. Reported on Minian 2.0.0, Python 3.12, latest xarray.

## Why xarray added this check

`safe_chunks=True` (xarray's default since the 2024.x rewrite of the zarr backend) guards against **silent data corruption**: when dask chunks straddle zarr-chunk boundaries, multiple parallel dask write tasks read-modify-write the *same* zarr chunk file and race — last writer wins, data is lost. The error is xarray correctly refusing to do an unsafe write.

## Fix

The first cut of this PR passed `safe_chunks=False` to silence the check. That suppresses the symptom but **disables the corruption guard** — and on the default `save_minian` path (no `chunks=` arg) there is no compensating rechunk, so the racy write actually happens. Confirmed by a round-trip test on the exact failing shape: `safe_chunks=False` writes data that does **not** equal the input.

This revision fixes the **root cause** instead: drop the stale `encoding['chunks']` from every variable before writing, so xarray derives zarr chunks from the live dask layout. Writes stay aligned, the `safe_chunks` guard stays at its safe default, and the error no longer fires.

```python
for v in ds.variables:
    ds[v].encoding.pop("chunks", None)
arr = ds.to_zarr(fp, compute=compute, mode=md)   # guard stays ON
```

## Verification

Round-trip byte-equality through the real `save_minian`, reproducing the issue's exact array (`C_lp`, stale `encoding['chunks']=(1,2000)`, live dask chunks `(1,1000)`) on xarray 2024.7.0:

| path | before (`safe_chunks=False`) | after (strip encoding) |
|---|---|---|
| default (no `chunks=`) | round-trip **NOT equal** (silent corruption) | round-trip **equal**, on-disk chunks `(1,1000)` |
| rechunker (`chunks=`) | — | round-trip **equal**, on-disk chunks `(1,1000)` |

Reported via miniscope/workshop-processing-and-analysis#3 (Bug 4).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Fix
+
+- `save_minian` no longer writes a stale `encoding['chunks']` carried over from a reloaded-and-rechunked array. When an array read back with `xr.open_zarr` was rechunked in memory and then saved again, that leftover tag could write chunk metadata disagreeing with the actual on-disk layout -- raising "Specified Zarr chunks ... would overlap multiple Dask chunks" on xarray >=2024, or silently corrupting the store on read if the chunk-alignment guard were disabled. `save_minian` now drops the stale tag and lets the on-disk grid follow the live dask layout. **Stock pipelines are unaffected** -- they only ever save freshly computed arrays. If you ran custom reload -> transform -> re-save code on xarray >=2024, re-verify those stores.
+
 ## v2.0.0 (2026-06-16)
 
 ### Highlight
--- a/minian/test/test_utilities.py
+++ b/minian/test/test_utilities.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 
+import dask.array as darr
 import numpy as np
 import xarray as xr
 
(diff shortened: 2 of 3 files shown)


sneakers-the-rat:
right, i get the basic problem, i'm just wondering how external rechunking + then using the `chunks` param works, would we then end up rechunking three times? or if we're simplifying how chunks work s.t. the in-memory chunks prevail, would we still want to have the chunks param here?

PR #347, opened 2026-08-25
fix: make ecos optional so minian installs on Python 3.13+

daharoni:
## Why

`ecos` publishes **no wheel for CPython 3.13 or newer**, and none at all for arm64 macOS. As a hard dependency that meant pip fell back to `ecos-2.0.14.tar.gz` and tried to compile C at install time on any newer interpreter.

That fails outright without a compiler toolchain. On managed machines it fails worse — a workshop participant on Windows + Python 3.13 hit this last week and saw a McAfee Artemis/Trojan alert, because the toolchain download itself tripped endpoint security. They spent a day treating it as an antivirus problem.

**Our CI never caught it.** GitHub runners ship MSVC and gcc, so the source build always succeeded. The 3.10–3.14 matrix proved the *code* runs on those versions; it never proved a *user* could install it. Worth keeping in mind for any dependency that's sdist-only on some interpreter.

## What changed

`ecos` becomes an opt-in extra (`pip install minian[ecos]`). `update_temporal_cvxpy` now resolves its solver at runtime:

```python
_CONE_SOLVERS = ("ECOS", "CLARABEL", "SCS")
```

ECOS stays **first**, so any environment that has it keeps producing exactly what it produced before — it's what the published parameter defaults were tuned against. Clarabel is the fallback: it's a required cvxpy dependency with wheels on every platform.

`ecos` was never imported anywhere in the codebase — it was only ever a cvxpy solver string — so nothing else moves.

### A sharp edge worth flagging

The solvers disagree on option spelling. ECOS and SCS take `max_iters`; **Clarabel takes `max_iter`**. cvxpy forwards solver options verbatim, so Clarabel raises `TypeError: unrecognized solver setting 'max_iters'` rather than ignoring it. The first cut of this change broke the fallback path for exactly that reason and only surfaced by running a real solve.

Mapped explicitly in `_MAX_ITER_OPT`, with `test_every_supported_solver_has_an_iteration_option` asserting the map covers `_CONE_SOLVERS` so a future addition can't skip it.

## Numerical impact

Same AR(2) problem, ECOS vs Clarabel:

| Metric | Result |
|---|---|
| calcium max abs diff | `6.1e-06` |
| calcium correlation | `1.000000` |
| spike correlation | `1.000000` |
| extra nonzeros under Clarabel | 87, all `< 4.6e-07` |
| median real spike magnitude | `0.031` |
| nonzero count above `1e-4` | identical (148 / 148) |

The only divergence is round-off dust sitting just above the default `zero_thres=1e-8`. Above `1e-6` the counts already agree.

## Test coverage

- `TestConeSolver` — preference order, caching, and the error when no supported solver is installed
- `TestTemporalUpdateSolvers` — drives a real AR(2) problem through **every** solver in `_CONE_SOLVERS` (skipping any not installed), plus an ECOS↔Clarabel equivalence check

CI: the default matrix now installs **without** ecos, so it exercises the fallback on every version and OS. A new `solver-ecos` job installs the extra on 3.12, asserts `cone_solver() == "ECOS"`, and runs the solver tests — so the ECOS path keeps dedicated coverage rather than losing it.

Locally: `ruff check` + `ruff format --check` clean repo-wide, 101 tests pass in `test_cnmf.py`. Built wheel metadata confirms it: `Requires-Dist: ecos>=2.0.7; extra == "ecos"`, and ecos absent from all 33 required deps.

`pdm.lock` regenerated with `--update-reuse`: ecos regrouped from `default` to `ecos`, **zero version changes** to anything else.

## Not included

- The `windows-latest × 3.13` matrix exclusion is left as-is. It's a deliberate cost tradeoff, and Linux 3.13 plus the new `solver-ecos` job cover the solver concern — but it does mean the exact configuration that broke still isn't in the matrix. Your call whether that's worth a job.
- No `requires-python` change. `>=3.10` was already accurate; the problem was a dependency, not the floor.

----
📚 Documentation preview 📚: https://minian--347.org.readthedocs.build/en/347/

### DIFF
--- a/.github/workflows/testandcov.yml
+++ b/.github/workflows/testandcov.yml
@@ -41,6 +41,35 @@ jobs:
           python -m pip install .
           minian data download --all
 
+  solver-ecos:
+    # The matrix above installs without `ecos`, so it only ever exercises the
+    # Clarabel fallback. ECOS is still the preferred solver wherever it is
+    # installed, and stays the one the published parameter defaults were tuned
+    # against, so it needs coverage of its own or a regression there would ship
+    # unnoticed. Scoped to the solver tests on one interpreter: ecos publishes
+    # no wheel past cp312, which is the whole reason it is optional now.
+    name: ECOS solver path py3.12
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install minian with the ecos extra
+        run: |
+          python -m pip install "pdm>=2.26.9"
+          pdm install --with test --with ecos
+      - name: Confirm ecos is the solver in use
+        run: |
+          pdm run python -c "
+          from minian.cnmf import cone_solver
+          got = cone_solver()
+          assert got == 'ECOS', f'expected ECOS with the extra installed, got {got}'
+          print('cone_solver() ->', got)
+          "
+      - name: Solver tests
+        run: pdm run pytest minian/test/test_cnmf.py -k "ConeSolver or TemporalUpdateSolvers" -p no:cacheprovider
+
   test-and-coverage:
     name: Test and Cov py ${{ matrix.python-version }}, ${{ matrix.os }}
     needs: prefetch-data
--- a/.github/workflows/testandcov.yml
+++ b/.github/workflows/testandcov.yml
@@ -42,12 +42,10 @@ jobs:
           minian data download --all
 
   solver-ecos:
-    # The matrix above installs without `ecos`, so it only ever exercises the
-    # Clarabel fallback. ECOS is still the preferred solver wherever it is
-    # installed, and stays the one the published parameter defaults were tuned
-    # against, so it needs coverage of its own or a regression there would ship
-    # unnoticed. Scoped to the solver tests on one interpreter: ecos publishes
-    # no wheel past cp312, which is the whole reason it is optional now.
+    # The matrix above installs without `ecos`, so it only exercises the Clarabel
+    # fallback. ECOS stays preferred wherever it is installed, so it needs its own
+    # coverage or a regression there ships unnoticed. One interpreter is enough:
+    # ecos has no wheel past cp312, which is why it is optional in the first place.
     name: ECOS solver path py3.12
     runs-on: ubuntu-latest
     steps:
@@ -59,15 +57,13 @@ jobs:
         run: |
           python -m pip install "pdm>=2.26.9"
           pdm install --with test --with ecos
-      - name: Confirm ecos is the solver in use
-        run: |
-          pdm run python -c "
-          from minian.cnmf import cone_solver
-          got = cone_solver()
-          assert got == 'ECOS', f'expected ECOS with the extra installed, got {got}'
-          print('cone_solver() ->', got)
-          "
+      # MINIAN_EXPECT_SOLVER makes the real-environment test assert that the
+      # extra actually won selection, rather than just resolving to something
+      # usable. Folded into the pytest run so the minian import stack (holoviews,
+      # datashader, dask, xarray) is paid once for this job instead of twice.
       - name: Solver tests
+        env:
+          MINIAN_EXPECT_SOLVER: ECOS
         run: pdm run pytest minian/test/test_cnmf.py -k "ConeSolver or TemporalUpdateSolvers" -p no:cacheprovider
 
   test-and-coverage:
--- a/docs/source/start_guide/install.rst
+++ b/docs/source/start_guide/install.rst
@@ -16,6 +16,8 @@ MiniAn is available on `PyPI <https://pypi.org/project/minian/>`_:
     python -m pip install minian
     # installation with optimized numba routines
     python -m pip install 'minian[numba]'
+    # add the ECOS solver, for results bit-identical to MiniAn <= 2.0.x
+    python -m pip install 'minian[ecos]'
 
 We recommend installing into a fresh virtual environment. A pip install does not
 include FFmpeg, so make sure it is on your ``PATH`` (see above).
--- a/minian/cnmf.py
+++ b/minian/cnmf.py
@@ -1186,52 +1186,42 @@ def update_temporal_block(
     return c, s, b, c0, g
 
 
-# Cone solvers the temporal update can use, most preferred first.
-#
-# ECOS stays first so that any environment which has it keeps producing exactly
-# the results it produced before -- it was a hard requirement for years and is
-# what the published parameter defaults were tuned against. It is optional now
-# (`pip install minian[ecos]`) because it ships no wheel past cp312, and none at
-# all for arm64 macOS, so requiring it made every newer interpreter compile it
-# from C source at install time. Clarabel is cvxpy's own default and one of its
-# required dependencies, so it is always available as a fallback.
-_CONE_SOLVERS = ("ECOS", "CLARABEL", "SCS")
+# Primary cone solvers for the temporal update, most preferred first. ECOS is
+# first for result-compatibility: it was required for years, so environments
+# that still have it must keep solving exactly as before. It is an optional
+# extra now (see CHANGELOG). SCS is deliberately absent -- it is the dedicated
+# retry solver below, and listing it here too would let it be chosen as primary
+# and then "fall back" to itself.
+_CONE_SOLVERS = ("ECOS", "CLARABEL")
 
-# cvxpy forwards solver options through verbatim, and these solvers spell their
-# iteration cap differently: ECOS and SCS take `max_iters`, Clarabel takes
-# `max_iter`. Clarabel rejects an unknown setting with TypeError rather than
-# ignoring it, so the name has to be mapped rather than passed blindly.
-_MAX_ITER_OPT = {"ECOS": "max_iters", "CLARABEL": "max_iter", "SCS": "max_iters"}
+# cvxpy forwards solver options verbatim and Clarabel spells its iteration cap
+# `max_iter`, rejecting the more common `max_iters` with TypeError rather than
+# ignoring it. Solvers absent from this map take the `max_iters` default.
+_MAX_ITER_OPT = {"CLARABEL": "max_iter"}
+_DEFAULT_MAX_ITER_OPT = "max_iters"
 
 
 @fct.lru_cache(maxsize=1)
 def cone_solver() -> str:
     """Name of the cone solver used for the temporal update.
 
-    Returns the first entry of :data:`_CONE_SOLVERS` that cvxpy reports as
-    installed. Cached, because :func:`cvxpy.installed_solvers` walks every
-    solver interface and this is called once per optimization problem.
-
-    Returns
-    -------
-    solver : str
-        A solver name accepted by :meth:`cvxpy.Problem.solve`.
+    Prefers ``ECOS`` when it is installed, otherwise ``CLARABEL``. Cached,
+    because :func:`cvxpy.installed_solvers` re-imports every absent solver on
+    each call and this runs once per optimization problem.
 
     Raises
     ------
     RuntimeError
-        If cvxpy reports none of them, which means the install is broken --
-        clarabel and scs are both required cvxpy dependencies.
+        If neither is installed. clarabel is a required cvxpy dependency, so
+        this means a broken environment rather than a missing extra.
     """
     installed = set(cvx.installed_solvers())
     for name in _CONE_SOLVERS:
         if name in installed:
             return name
     raise RuntimeError(
-        "None of the supported cone solvers "
-        f"({', '.join(_CONE_SOLVERS)}) are installed. cvxpy requires clarabel "
-        "and scs, so this usually means a broken environment; reinstall with "
-        "`pip install --force-reinstall cvxpy`."
+        f"None of the supported cone solvers ({', '.join(_CONE_SOLVERS)}) are "
+        "installed; reinstall with `pip install --force-reinstall cvxpy`."
     )
 
 
@@ -1407,11 +1397,11 @@ def update_temporal_cvxpy(
         )
         prob = cvx.Problem(obj, cons)
         try:
-            _solver = cone_solver()
+            solver = cone_solver()
             _ = prob.solve(
-                solver=_solver,
+                solver=solver,
                 warm_start=warm_start,
-                **{_MAX_ITER_OPT[_solver]: max_iters},
+                **{_MAX_ITER_OPT.get(solver, _DEFAULT_MAX_ITER_OPT): max_iters},
             )
             if prob.status in ["infeasible", "unbounded", None]:
                 raise ValueError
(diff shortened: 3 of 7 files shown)


.github/workflows/testandcov.yml
```diff
@@ -41,6 +41,31 @@ jobs:
...
+          python -m pip install "pdm>=2.26.9"
+          pdm install --with test --with ecos
+      # MINIAN_EXPECT_SOLVER makes the real-environment test assert that the
+      # extra actually won selection, rather than just resolving to something
+      # usable. Folded into the pytest run so the minian import stack (holoviews,
+      # datashader, dask, xarray) is paid once for this job instead of twice.
+      - name: Solver tests
+        env:
+          MINIAN_EXPECT_SOLVER: ECOS
+        run: pdm run pytest minian/test/test_cnmf.py -k "ConeSolver or TemporalUpdateSolvers" -p no:cacheprovider
```
sneakers-the-rat, line 67: would be cleaner and easier to understand to use pytest marks for the things that we want to test with ecos, finding tests by name in CI is fragile and doesn't scale well.

docs/source/start_guide/install.rst
```diff
@@ -16,6 +16,8 @@ MiniAn is available on `PyPI <https://pypi.org/project/minian/>`_:
     python -m pip install minian
     # installation with optimized numba routines
     python -m pip install 'minian[numba]'
+    # add the ECOS solver, for results bit-identical to MiniAn <= 2.0.x
+    python -m pip install 'minian[ecos]'
```
sneakers-the-rat, line 20: i'm not sure if downstream users would understand when/why they would want to do this

minian/test/test_cnmf.py
```diff
@@ -606,3 +610,172 @@ def test_empty_membership_returns_empty_diag(self):
...
+
+    @pytest.mark.parametrize(
+        ("installed", "expected"),
+        [
+            (["ECOS", "CLARABEL", "SCS"], "ECOS"),
+            (["CLARABEL", "SCS"], "CLARABEL"),
+            # Reported order must not matter, only _CONE_SOLVERS order.
+            (["SCS", "CLARABEL"], "CLARABEL"),
+            (["OSQP", "CLARABEL", "ECOS"], "ECOS"),
+        ],
```
sneakers-the-rat, line 649: what are these other solvers ("SCS", "OSQP") and why are they present if they are never used?

minian/test/test_cnmf.py
```diff
@@ -606,3 +610,172 @@ def test_empty_membership_returns_empty_diag(self):
...
+    )
+    def test_picks_first_available_in_preference_order(
+        self, monkeypatch, installed: list[str], expected: str
+    ):
+        monkeypatch.setattr(cnmf.cvx, "installed_solvers", lambda: installed)
+        assert cone_solver() == expected
+
+    @pytest.mark.parametrize("installed", [["OSQP"], ["SCS"], []])
+    def test_raises_when_no_primary_solver_is_installed(self, monkeypatch, installed: list[str]):
+        # clarabel is a required cvxpy dep, so none of these can happen in a
```
sneakers-the-rat, line 659: then why would we have this test?

minian/cnmf.py
```diff
@@ -1184,6 +1186,45 @@ def update_temporal_block(
...
+_CONE_SOLVERS = ("ECOS", "CLARABEL")
+
+# cvxpy forwards solver options verbatim and Clarabel spells its iteration cap
+# `max_iter`, rejecting the more common `max_iters` with TypeError rather than
+# ignoring it. Solvers absent from this map take the `max_iters` default.
+_MAX_ITER_OPT = {"CLARABEL": "max_iter"}
+_DEFAULT_MAX_ITER_OPT = "max_iters"
+
+
+@fct.lru_cache(maxsize=1)
```
sneakers-the-rat, line 1204:
```suggestion
@fct.cache
```

the function takes no arguments, so there is always only one cache entry, and `cache` is faster

minian/test/test_cnmf.py
```diff
@@ -606,3 +610,172 @@ def test_empty_membership_returns_empty_diag(self):
...
+        calls = []
+
+        def _installed() -> list[str]:
+            calls.append(1)
+            return ["CLARABEL"]
+
+        monkeypatch.setattr(cnmf.cvx, "installed_solvers", _installed)
+        assert cone_solver() == "CLARABEL"
+        assert cone_solver() == "CLARABEL"
+        assert len(calls) == 1
```
sneakers-the-rat, line 678:
```suggestion
```

we don't need to test stdlib

minian/test/test_cnmf.py
```diff
@@ -606,3 +610,172 @@ def test_empty_membership_returns_empty_diag(self):
...
+        # No monkeypatching: whatever this install actually has must be
+        # something cvxpy will accept. Catches a preference list that has
+        # drifted from the solvers cvxpy ships.
+        got = cone_solver()
+        assert got in set(cnmf.cvx.installed_solvers())
+        # CI's solver-ecos job sets this to prove the extra really wins
+        # selection, not merely that selection produced something usable.
+        expected = os.environ.get("MINIAN_EXPECT_SOLVER")
+        if expected:
+            assert got == expected, f"expected {expected}, selected {got}"
```
sneakers-the-rat, line 690: no idea what this does - it should just be impossible to give an impossible value

minian/test/test_cnmf.py
```diff
@@ -606,3 +610,172 @@ def test_empty_membership_returns_empty_diag(self):
...
+
+
+# ---------------------------------------------------------------------------
+# update_temporal_cvxpy solver compatibility
+# ---------------------------------------------------------------------------
+
+
+def _ar2_problem(
```
sneakers-the-rat, line 698: surely we have data synthesis functions already....

minian/test/test_cnmf.py
```diff
@@ -606,3 +610,172 @@ def test_empty_membership_returns_empty_diag(self):
...
+    Regression guard: cvxpy forwards solver options verbatim, and Clarabel
+    spells its iteration cap ``max_iter`` where ECOS and SCS use ``max_iters``.
+    Passing the wrong one raises TypeError instead of being ignored, so routing
+    the solve through a non-ECOS solver broke until the name was mapped.
+    """
+
+    @pytest.mark.parametrize("solver", _CONE_SOLVERS)
+    def test_solve_succeeds_under_each_installed_solver(self, monkeypatch, solver):
+        if solver not in cnmf.cvx.installed_solvers():
+            pytest.skip(f"{solver} not installed")
```
sneakers-the-rat, line 730: when would we want to do this rather than failing? shouldn't we expect the testing environment to always have the conditions that we need to test the code?

minian/test/test_cnmf.py
```diff
@@ -606,3 +610,172 @@ def test_empty_membership_returns_empty_diag(self):
...
+        y, g, sn = _ar2_problem()
+        monkeypatch.setattr(cnmf.cvx, "installed_solvers", lambda: [solver])
```
sneakers-the-rat, line 732: if this is the only way to set the solver, we can tell the API is badly designed. `get_solver`/`set_solver` if this is something that we want to be configurable.

minian/test/test_cnmf.py
```diff
@@ -606,3 +610,172 @@ def test_empty_membership_returns_empty_diag(self):
...
+            sn,
+            sparse_penal=1.0,
+            max_iters=200,
+            use_cons=False,
+            scs_fallback=False,
+            c_last=None,
+            zero_thres=1e-8,
+            warm_start=False,
+        )
+        assert c.shape == y.shape
```
sneakers-the-rat, line 745: so this seems like just bare minimum sanity tests. if what we want to do is support and test against multiple solvers that shoudl be considered equivalent to a downstream user, we want to parameterize _all_ the tests that use a solver, rather than creating a few weak duplicate tests

minian/test/test_cnmf.py
```diff
@@ -606,3 +610,172 @@ def test_empty_membership_returns_empty_diag(self):
...
+        # An all-zero return is how the function signals total solver failure.
+        assert c.max() > 0, f"{solver} produced an all-zero solution"
+
+    def test_clarabel_matches_ecos_on_the_same_problem(self, monkeypatch):
+        # The guarantee behind making ecos optional: dropping it must not
+        # change the answer. Tiny differences survive near zero_thres, so
+        # compare the recovered traces rather than exact spike support.
+        installed = cnmf.cvx.installed_solvers()
+        if not {"ECOS", "CLARABEL"} <= set(installed):
+            pytest.skip("needs both ECOS and CLARABEL installed")
```
sneakers-the-rat, line 757: again, fail here

.github/workflows/testandcov.yml
sneakers-the-rat: commented below, but the parameterization to test the different solvers should be within the tests, not within CI - the testing install should require ecos installation if it is a supported solver.

minian/test/test_cnmf.py
```diff
@@ -606,3 +610,172 @@ def test_empty_membership_returns_empty_diag(self):
...
+        kw = {
+            "sparse_penal": 1.0,
+            "max_iters": 200,
+            "use_cons": False,
+            "scs_fallback": False,
+            "c_last": None,
+            "zero_thres": 1e-8,
+            "warm_start": False,
+        }
+        res = {}
```
sneakers-the-rat, line 768: since there is an absolute shitload of mutation-in-place in this package, probably would be a tighter test to create these within the loop rather than before it

minian/test/test_cnmf.py
```diff
@@ -606,3 +610,172 @@ def test_empty_membership_returns_empty_diag(self):
...
+        for name in ("ECOS", "CLARABEL"):
+            monkeypatch.setattr(cnmf.cvx, "installed_solvers", lambda n=name: [n])
+            cone_solver.cache_clear()
+            res[name] = cnmf.update_temporal_cvxpy(y.copy(), g, sn, **kw)
+
+        c_ecos, c_clara = res["ECOS"][0], res["CLARABEL"][0]
+        assert np.corrcoef(c_ecos.ravel(), c_clara.ravel())[0, 1] > 0.9999
```
sneakers-the-rat, line 775: why not `allclose`?

minian/test/test_cnmf.py
```diff
@@ -606,3 +610,172 @@ def test_empty_membership_returns_empty_diag(self):
...
+        assert np.abs(c_ecos - c_clara).max() < 1e-3
+
+        # Spike support agrees once round-off dust just above zero_thres is
+        # excluded -- below 1e-6 the two solvers disagree on near-zero values.
+        s_ecos, s_clara = res["ECOS"][1], res["CLARABEL"][1]
+        assert (s_ecos > 1e-4).sum() == (s_clara > 1e-4).sum()
```
sneakers-the-rat, line 781:
```suggestion
        assert np.allclose(s_ecos, s_clara, atol=1e-5)
```

we care about more than just the number of values that are above some threshold - we care that the values are near-identical.

minian/cnmf.py
```diff
@@ -1184,6 +1186,45 @@ def update_temporal_block(
     return c, s, b, c0, g
 
 
+# Primary cone solvers for the temporal update, most preferred first. ECOS is
+# first for result-compatibility: it was required for years, so environments
+# that still have it must keep solving exactly as before. It is an optional
+# extra now (see CHANGELOG). SCS is deliberately absent -- it is the dedicated
+# retry solver below, and listing it here too would let it be chosen as primary
+# and then "fall back" to itself.
```
sneakers-the-rat, line 1194: why would we want a retry solver at all?

minian/cnmf.py
```diff
@@ -1184,6 +1186,45 @@ def update_temporal_block(
...
+_CONE_SOLVERS = ("ECOS", "CLARABEL")
+
+# cvxpy forwards solver options verbatim and Clarabel spells its iteration cap
+# `max_iter`, rejecting the more common `max_iters` with TypeError rather than
+# ignoring it. Solvers absent from this map take the `max_iters` default.
+_MAX_ITER_OPT = {"CLARABEL": "max_iter"}
+_DEFAULT_MAX_ITER_OPT = "max_iters"
+
```
sneakers-the-rat, line 1202: super strange and brittle. if we need to have some compatibility mapping between different backends, then we specify our spelling and a map to the different backend spellings. something like

```python
_BACKEND_KWARGS = {
  "CLARABEL": {
    "max_iters": "max_iter",
  },
}
```

and then when calling clarabel, something like this

```python
kwarg_map = _BACKEND_KWARGS["CLARABEL"]
kwargs = {kwarg_map.get(k, k): v for k, v in kwargs.items()}
```

pyproject.toml
```diff
@@ -11,7 +11,6 @@ dependencies = [
     "dask>=2025.1.0",
     "datashader>=0.12.1",
     "distributed>=2025.1.0",
-    "ecos>=2.0.7",
     "ffmpeg-python>=0.2.0",
```
sneakers-the-rat, line 14: the way i would probably do this is leave it required for python <=3.12 with a platform flag, but whatevzies

pyproject.toml
```diff
@@ -82,6 +81,11 @@ minian = "minian.cli:main"
 minian-install = "minian.install:main"
 
 [project.optional-dependencies]
+# Opt-in: keeps the temporal update on ECOS for results bit-identical to
+# MiniAn <= 2.0.x. Not required, because ecos has no wheel past cp312.
```
sneakers-the-rat, line 85: this goes in the docs. by virtue of being in `optional-dependencies` we already know that it's not required and opt-in.

sneakers-the-rat (commented):
So the simplest way to do this would just be to make the solver a user-configurable parameter and drop `ecos` entirely, and then if someone wants to use `ecos` they can install it themselves and use it. this is alright, but notice how we need to do a ton of monkeypatching to set the solver - it would be cleaner for us to have get/set solver already, so by just having a get/set solver method in the first place, we simplify both the changes here *and* create more explicit flexibility for users.

the biggest missing piece here is any information about when one would want to use either of these solvers: is `clarabel` slower? faster? more accurate? less accurate? would someone want to control which solver they use? when? what are the tradeoffs? being bit-identical with <2.0 is not informative, what people want from an analysis package is that the numbers are as accurate as they can be as fast as they can be computed, in that order of priority, so numerical divergence is *good* if we are improving accuracy (assuming we have published versions with changelogs so someone can pin an old version, or set the solver themselves as above). 

if we are keeping an optional installation candidate, the docs need to answer "when would i want to do that." and they should also describe the above tradeoffs between the available options if we want to officially support both. 

i have no idea what the "fallback solver" thing that's mentioned in the comments is, but having a fallback solver is a big time code smell: in what circumstances would we want a fallback solver rather than resolving the conditions that led us to need a fallback in the first place?

sneakers-the-rat:
> median real spike magnitude 	0.031

i'm not sure what the absolute value of the spike values is, but this seems like it could be relatively large (e.g. if the absolute spike values are ~0.1)

